PR #187, opened 2015-01-17
Update atom-beautify.cson

laszlof:
FIx deprecated selectors in latest version of Atom.


emcek:
should be:
```.atom-workspace .atom-text-editor:not(.mini)':```

Glavin001:
Looks like #193 uses correct selectors. Thanks for your PR though!

dinosaurfiles:
screenshot from 2015-01-31 23 57 29
screenshot from 2015-01-31 23 57 45
Doesnt work on the latest 0.21.2 update tho

PR #193, opened 2015-01-26
updating selectors to the new atom API

tcarlsen:
atom-beautify
  menus/atom-beautify.cson
    Use the `atom-workspace` tag instead of the `workspace` class.
    Use the `atom-text-editor` tag instead of the `editor` class.


Glavin001:
Published to v0.21.3.
Thank you for contributing!

tcarlsen:
my turn I guess ;)

PR #251, opened 2015-03-25
Add check for Windows and use windows specific 'mv' command

rrushton:
I don't have a windows machine to test on, but its a pretty simple command so it should work. Hopefully someone with a Windows machine can test this.


Glavin001:
I also do not have a Windows computer. We can let the user's test this. :+1:

Glavin001:
Published to v.23.1

PR #263, opened 2015-03-31
php is not necessary

MGAio:


jaswsinc:
Just wanted to say that I agree with this PR.

The `php` prefix is actually causing problems for me (on OS X), because `php` refers to `/usr/bin/php` instead of to `/usr/bin/env php` (i.e., `/usr/local/bin/php`).

In any case, the leading `php` being removed leaves this up to the shebang line in `php-cs-fixer`, which is what _should_ occur IMO :-)

Glavin001:
Looks good to me. Thanks!

Glavin001:
Published to v0.23.2

PR #275, opened 2015-04-11
replacing rbeautify with rubocop

jney:


miketheman:
Something worth considering is that since this uses rubocop's `-a` flag to autocorrect, will it load a custom .rubocop.yml file from <some directory> and if so, which one?

Glavin001:
I do not want to break support for current users using `rbeautify`. How about this supports both? If `rbeautify_path` is set then use rbeautify, or if `rubocop_path` is set then use rubocop.

jney:
> How about this supports both? 

@Glavin001, i thought about it. The problem was `isStdout` is set at the language inititialization, not at the beautifier launch, it involves that you must know at the beginning which beautifier you would use (whatever how many are declared).
Replacing was very simpler than supporting both.

Glavin001:
> The problem was isStdout is set at the language inititialization, not at the beautifier launch, it involves that you must know at the beginning which beautifier you would use (whatever how many are declared).

Very good point. I think I will end up merging this and migrating to rubocop. However, I do want to wait until I have redesigned Atom Beautify's internals. The new structure should allow supporting both and make adding beautifiers much easier. I don't want to break compatibility for any current users, at least not just yet -- although we did have to do this in the past before, and it caused many confusing issues to be created.

I just created #282 and once that is complete we will have support for *both* rubocop and rbeautify (legacy). We can potentially add deprecation warnings, too.

Glavin001:
Almost done with #282! Then we can have both rubocop and rbeautify and it will be much easier to add more beautifiers, too!

Glavin001:
This will be in the next release.

You can select your preferred beautifier, if it is not default, for each language in the Atom Beautify package settings:
image

Issue #267, opened 2015-04-05
Some calls are deprecated since Atom 0.189.0


d4k0:
Since the update to Atom 0.189.0, these messages appear:

![Deprecated calls](https://i.imgur.com/TMsona6.png)

It would be nice if this could be fixed.

Glavin001:
Please update Atom Beautify before creating new issues.

I believe these have been fixed for a while , #193 , #172 , #195 , etc. 
Also see https://github.com/Glavin001/atom-beautify/blob/master/package.json#L72-L85 which *does* use `activationCommands` instead of `activationEvents`.

Thanks!

d4k0:
Of course I updated to the latest version (0.23.2) of Atom Beautify before creating this issue, but these deprecated messages are still there. 

BTW: Atom Beautify is not the only package that shows these messages since the update to Atom 0.189.0. I have 4 other packages that also use some deprecated calls.

Glavin001:
This is strange. I will have to look into this closer after my midterms. Thanks for confirming.

d4k0:
No problem. You don't need to hurry as Atom Beautify still seems to work without a problem.

Atom 0.189.0 seems to have changed quite a few things regarding deprecated calls (you can see this when you look at the commits to the Atom repository, but it's not mentioned in the changelog) as the other packages all mention different calls (or directory names for themes) that shouldn't be used anymore.

Glavin001:
I am unable to reproduce your results above in the Deprecation Cop. Could you provide some information on how you were able to achieve those deprecation warnings? I am currently opening Atom and beautifying files which should trigger those warnings.

Also if you could run `Beautify : Help Debug Editor` and paste contents here, that may be helpful. 
Thanks!

d4k0:
I just tried it with a new profile with only Atom Beautify installed and this is how it looks:

![New profile with Atom Beautify](https://i.imgur.com/jsFSGbw.png)

As you can see Atom now somehow only shows one of the deprecation warnings here. It seems that maybe some of my other packages affect Atom Beautify (I already could observe this with other packages that complain about the stylesheet folder of another package), but so far I haven't had any luck to find out which one of the packages is the culprit (I will do some more tests).

Regarding the deprecation warnings: I get them when I start Atom (no file has to be loaded):

![Deprecation warnings](https://i.imgur.com/5dJGH5Z.png)

[...]

PR #279, opened 2015-04-16
Updates context-menu to use new api

wanton:
This removes the warning "ContextMenuManager::add has changed to take a single object as its argument. Please see https://atom.io/docs/api/latest/ContextMenuManager for more info." which is one of the messages described in #267


Glavin001:
Thanks for contributing!

Glavin001:
Published to v0.24.1

PR #330, opened 2015-05-07
Do not inherit env from the parent process (Atom).

fd:
Inheriting the environment messes with the environment from the shell.


Glavin001:
I just tested this and it actually broke my latest tests that are applicable. So I am not going to merge this. Thanks for contributing.

PR #348, opened 2015-05-19
adds .js to the extensions supported by the jsx language

### LINKED ISSUE #144: Support for JSX syntax (ReactJS)
Hi.
I think, it's hard to implement, but would be useful for a lot of people.
Have you this in plans?

luisrudge:
fixes #144 for good

### DIFF
--- a/src/languages/jsx.coffee
+++ b/src/languages/jsx.coffee
@@ -16,7 +16,8 @@ module.exports = {
     Supported extensions
     ###
     extensions: [
-        "jsx"
+        "jsx",
+        "js"
     ]
 
 }


Glavin001:
This actually will not fix the problem you are having ( #144 ), however I am merging it anyway because it will not change anything. 
While the addition of the `js` extension is not necessary, it will not break anything because the priority of the grammar supersedes the extensions ( see https://github.com/Glavin001/atom-beautify/blob/master/src/languages/index.coffee#L86 ). The `JavaScript` grammar will of course be default in Atom for `.js` unless changed by user.

Furthermore, the real bug is #353. Basically, `extensions` was not working, at all. I had a couple typos. It was actually only using the `grammars`, for which both `JavaScript (JSX)` and `JSX` are already supported.

I will be publishing fix for #353 soon, however given any `extensions` are effectively useless in the currently published version, I wonder: did you test this pull request? Did it work? If so, then it was something else, such as restarting Atom or updating Atom Beautify, etc, that actually is making it work for you now, as the grammars `JavaScript (JSX)` and `JSX` were already added and usable under language `JSX`.

Thank you for your contributions, regardless.

Glavin001:
Published to v0.27.4

luisrudge:
Thanks for the explanation.

PR #357, opened 2015-05-22
Enable typescript-formatter defaults.

yourpalal:
These defaults were in place until 25e2a754d58. Without them, there is an unfortunate lack of whitespace (IMO).


Glavin001:
Thank you! I was wondering why it was causing problems with the tests after merging that Pull Request. I thought maybe it was change in the TypeScript Formatter source code when I updated the dependency.

I'll review soon and merge ASAP.

Glavin001:
Thanks for also taking care of the tests!

Glavin001:
Published to v0.27.6

yourpalal:
Thanks for the quick merge! :)

PR #382, opened 2015-06-01
Fixing faux possessive 'it is' -> 'its', and some small language fixes

psy-q:
Just spotted that in a notification about a missing beautifier, so I thought I'd correct it.


Glavin001:
Thanks!

PR #429, opened 2015-06-28
Improve README with Atom Beautify package settings

blelump:


blelump:
@Glavin001 , I see. I've tried to find in package settings a field related to ERB beautifier selection, however I couldn't find any. Besides that, there're plenty of options which I am not sure, I need them (or anyone else), so at least I would group these options somehow. How are they built? Perhaps I would try to play with them. 

The context menu misunderstanding is probably due to different platform. You're on OS X (Atom -> Open Your Config) and I'm on Linux (Edit -> Open Your Config).

Glavin001:
>  I've tried to find in package settings a field related to ERB beautifier selection, however I couldn't find any.

You can change the ERB default beautifier with `Language Config - ERB - Default Beautifier`: https://github.com/Glavin001/atom-beautify/blob/master/docs/options.md#language-config---erb---default-beautifier

> Besides that, there're plenty of options which I am not sure, I need them (or anyone else), so at least I would group these options somehow. How are they built? Perhaps I would try to play with them.

The documented options here: https://github.com/Glavin001/atom-beautify/blob/master/docs/options.md
Are generated by https://github.com/Glavin001/atom-beautify/blob/master/docs/index.coffee
Which uses the options generated by `buildOptionsForBeautifiers` here: https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/index.coffee#L92-L273

> The context menu misunderstanding is probably due to different platform. You're on OS X (Atom -> Open Your Config) and I'm on Linux (Edit -> Open Your Config).

Yup. I think the platform agnostic way is to use the Atom command `Application: Open your Config`

Glavin001:
This README change is not really in the direction that I want to go with documentation. I do agree, there are too many options, however I still encourage users to read the documentation (link provided in current README already) and then set in their Atom Beautify package settings, and in their `.jsbeautifyrc` file for project-specific settings. This is all indicated in the old / current README, so I am going to close this.

I do however, want to look into improvements for displaying the settings in the package settings view. I wondering if groups are supported?

PR #436, opened 2015-06-30
add language scopes to atom language config gets

tcl16:
Atom's config is not fully supported at the moment. Config like the following is currently ignored:

```cson
"*":
  welcome:
    showOnStartup: false
  linter:
    lintOnChange: false
    clearOnChange: true
  core: {}
".shell.source":
  editor:
    tabLength: 4
".python.source":
  editor:
    tabLength: 4
".java.source":
  editor:
    tabLength: 4
".gfm.source":
  editor:
    showInvisibles: false
    softWrap: true
    tabLength: 4
```

atom-beautify currently ignores the language specific tabLength and other settings and uses the default tabLength of 2 in this case. This commit fixes that issue by specifying the scope in the language files.

### DIFF
--- a/src/languages/css.coffee
+++ b/src/languages/css.coffee
@@ -1,6 +1,7 @@
 # Get Atom defaults
-tabLength = atom?.config.get('editor.tabLength') ? 4
-softTabs = atom?.config.get('editor.softTabs') ? true
+scope = ['source.css']
+tabLength = atom?.config.get('editor.tabLength', scope: scope) ? 4
+softTabs = atom?.config.get('editor.softTabs', scope: scope) ? true
 defaultIndentSize = (if softTabs then tabLength else 1)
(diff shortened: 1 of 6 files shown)


Glavin001:
I did not know that Atom had a language scope for those editor settings! This is excellent, thank you!

tcl16:
So I now realize that it was not just these languages that did not support this. If you would like I can try to expand support for this style of configuration to the other languages as well.

Glavin001:
That would be great, thanks! This is definitely something I wanted to support, I just did not know that Atom had scoping for settings like this.

Glavin001:
@tcl16 I am going to merge this now. Feel free to submit another Pull Request for those other languages you wish to add. Thank you for contributing!

Glavin001:
Published to v0.28.9

PR #464, opened 2015-07-17
Beautify Swig templates with handlebars instead of markup

### LINKED ISSUE #463: HTML5 void elements in Swig
If I have markup like this in a Swig file:
```html
<div>
  <img>
  <img>
</div>
```

The result will be:
```html
<div>
  <img>
    <img></div>
```

It seems that it cannot handle the HTML5 void elements correctly. However, If I manually change the language to HTML, everything works and the result is the expected.

If I understood correctly, the Swig formatter uses the same formatter as HTML, ie. there is no special formatter for Swig. What could cause this difference?

ttsirkia:
Fixes #463


ttsirkia:
OK, no idea what the error message of AppVeyor means.

Glavin001:
@prettydiff, you get the final say, if `Swig` should be under `markup` or `handlebars`. Thanks!

Glavin001:
AppVeyor is dumb sometimes. I mostly follow Travis CI since it's much more reliable (read: "not Windows").

Thanks for contributing!

Glavin001:
Published to v0.28.8

PR #476, opened 2015-07-23
:bug: fix the apparent race condition in moving to line 1 and display…

### LINKED ISSUE #472: Beautify on save moves the cursor at the top of the file
Here are my debug logs like you asked in #338 

[EDIT]: Move debug info into a Gist

https://gist.github.com/sebz/54411974e251070926c7

kelvix:
…ing text

Essentially a copy and paste from [setTimer](https://github.com/Glavin001/atom-beautify/blob/a2efade4b3eaac5d61c1da0ec17eb396e231b672/src/beautify.coffee#L127-L136) need to Google that next...

Appears to fix #472 my test devices include OS X 10.11 and Fedora, can test on Windows tomorrow don't feel like turning that guy on tonight.

### DIFF
--- a/src/beautify.coffee
+++ b/src/beautify.coffee
@@ -368,7 +368,7 @@ debug = () ->
     addInfo('Final Options', "\n" +
     "Final combined options that are used\n" +
     "```json\n#{JSON.stringify(finalOptions, undefined, 4)}\n```")
-    
+
     addInfo('Package Settings', "\n" +
     "The raw package settings options\n" +
     "```json\n#{JSON.stringify(atom.config.get('atom-beautify'), undefined, 4)}\n```")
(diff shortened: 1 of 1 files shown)


nikmartin:
This does work for me. The screen flashes , but I'm quickly brought back to the cursor location before save. I just patched v0.28.8 with the changes and turned language_js_beautify_on_save back on, and it worked.

kelvix:
The screen flash is, what I believe to be at least, the text being removed and added. The beautify process if I step through it at one point removes all the text from the editor. Once completed brings the text back to the editor. Going to play with it some more to see if I can keep that from happening. Another thing I just noticed if you save really fast say 5 or 6 times we do eventually end up at line 1 again.

kelvix:
So I spent more time of this than I would have liked before trying to troubleshoot [tree-view](https://github.com/atom/tree-view). Would you mind testing beautify on save with no directories in your tree-view?

What I noticed was that I cloned their latest master branch and linked it to my atom development instance. Tried performing the beautify on save and it kinda just worked... leaving me with little to debug... If things work out for you as they did for me I can go out and submit an issue to tree-view or atom. To me it feels more like an Atom issue and not so much tree-view

nikmartin:
Well, it's a bit more inconsistent than that, but not worth bothering with possibly. CTRL+S beautifies the file perfectly with no flash or glitches with the tree-view open, *as long as the file being beautified isn't visible in the tree view*

So, I tested your suggestion and it worked fine, then I added a project folder back, displaying the tree view, but the folder the file I was saving was in a folder that was collapsed, and it beautified fine. When I opened the folder the file is in, the screen flashes on save.

Glavin001:
Published to v0.28.9

Issue #479, opened 2015-07-24
perltidy error when perltidy_profile is an empty string


Linghui:
perltidy on pl file makes the file gone empty.
perltidy package works ok itself, but beautify does, so I thought the problem is on beautify side.

Glavin001:
Could you run Atom command `Atom Beautify - Help Debug Editor` and paste the results into a Gist and link here please? Let's try to debug this problem.

kelvix:
@Glavin001 I'll be creeping around your issues list because I find this too fun for the time being... In any case seems to getting caused by [perltidy.coffee](https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/perltidy.coffee#L26). options.perltidy_profile? actually returns an empty string instead of null or undefined causing a stderr of

    "usage: -pro=filename or --profile=filename, no spaces"

Sounds like there would need to be a check for an empty string i.e options.perltidy_profile?.length I can branch and issue a pull request if you'd like.

Linghui:
@Glavin001  I run 'Atom Beautify - Help Debug Editor' , it just open 5 empty files named Beautify, -, Help ,Debug and Editor.

kelvix:
@Linghui I think you used the terminal to perform that command. Not sure if the terminal has an equivalent at the moment for that command. Glavin001 was probably after the results of running that command from the Command Palette. The Command Palette can be found at Packages -> Command Palette -> Toggle. The command in Command Palette in Atom 1.0.2 appears to actually be `Atom Beautify: Help Debug Editor`. 

In addition the file you're debugging on has to have a name.

Linghui:
@kelvix Thank you for the information :) 
@Glavin001 Here is the debug info, https://gist.github.com/Linghui/158a776bcdae006322f7, hope this would help.

Glavin001:
@kelvix I think you are on the right track in your comment https://github.com/Glavin001/atom-beautify/issues/479#issuecomment-124783739 
I would love to receive a Pull Request :smiley: 
Thanks!

Glavin001:
Published to v0.28.9

PR #486, opened 2015-07-29
bug/perltidy

kelvix:
should cover #479 by not passing the --profile option when an empty string is passed as its argument.


Glavin001:
Published to v0.28.9

Issue #482, opened 2015-07-25
Uncaught TypeError: Cannot read property 'name' of null


kelvix:
Not sure if this is intentional or not but if the file we try to debug on does not have a name yet we get this exception.

1. Attempt to use Atom Beautiy: Help Debug Editor in a new file that has not been saved. 

**Atom Version**: 1.0.2
**System**: "Fedora release
**Thrown From**: [atom-beautify](https://github.com/Glavin001/atom-beautify) package, v0.28.8

### Stack Trace

Uncaught TypeError: Cannot read property 'name' of null

```
At /home/jgarcia/.atom/packages/atom-beautify/src/beautify.coffee:325

TypeError: Cannot read property 'name' of null
  at atom-workspace.debug (/home/jgarcia/.atom/packages/atom-beautify/src/beautify.coffee:325:45)
  at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/usr/share/atom/resources/app.asar/src/command-registry.js:241:29)
  at /usr/share/atom/resources/app.asar/src/command-registry.js:3:61
  at CommandPaletteView.module.exports.CommandPaletteView.confirmed (/usr/share/atom/resources/app.asar/node_modules/command-palette/lib/command-palette-view.js:159:32)
  at CommandPaletteView.module.exports.SelectListView.confirmSelection (/usr/share/atom/resources/app.asar/node_modules/atom-space-pen-views/lib/select-list-view.js:338:21)
  at space-pen-li.<anonymous> (/usr/share/atom/resources/app.asar/node_modules/atom-space-pen-views/lib/select-list-view.js:139:19)
  at HTMLOListElement.jQuery.event.dispatch (/usr/share/atom/resources/app.asar/node_modules/jquery/dist/jquery.js:4435:9)
  at HTMLOListElement.elemData.handle (/usr/share/atom/resources/app.asar/node_modules/jquery/dist/jquery.js:4121:28)

```

### Commands

```
     -4:35.7.0 core:backspace (atom-text-editor.editor.mini.is-focused)
     -4:34.9.0 editor:consolidate-selections (atom-text-editor.editor.mini.is-focused)
     -4:34.9.0 core:cancel (atom-text-editor.editor.mini.is-focused)
     -4:28.2.0 command-palette:toggle (a.stack-toggle)
  2x -4:27.3.0 core:paste (atom-text-editor.editor.mini.is-focused)
     -4:11.7.0 command-palette:toggle (a.stack-toggle)
[...]

PR #487, opened 2015-07-30
bug/nullgrammar

kelvix:
Attempts to handle #482. Don't know how I feel about this one. In order to handle this language cannot be null. The nested if statement is used to display an appropriate UI either we tried debugging a Null Grammar or a grammar we don't support. We also arbitrarily return at the end of the if statement or else we still get our big red flag. 

I'm open to suggestions.

### DIFF
--- a/src/beautify.coffee
+++ b/src/beautify.coffee
@@ -368,7 +368,7 @@ debug = () ->
     addInfo('Final Options', "\n" +
     "Final combined options that are used\n" +
     "```json\n#{JSON.stringify(finalOptions, undefined, 4)}\n```")
-    
+
     addInfo('Package Settings', "\n" +
     "The raw package settings options\n" +
     "```json\n#{JSON.stringify(atom.config.get('atom-beautify'), undefined, 4)}\n```")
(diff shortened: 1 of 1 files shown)


Glavin001:
Yeah I actually do want to see the resulting debugging contents, not return another error notification to the user that they have to figure out.

I would prefer simply:

```coffeescript
addInfo('Original File Language', language.name)
```

Be changed into:

```coffeescript
addInfo('Original File Language', language?.name)
```

Which is in JavaScript:

```javascript
addInfo('Original File Language', typeof language !== "undefined" && language !== null ? language.name : void 0);
```

This means it works smoothly and displays `undefined` for the `Original File Language` which is something I could see in the debugs and would potentially be useful information.

Thank you for contributing! If you would like to create a new Pull Request with that little change, I'll merge it.

kelvix:
That was the first thing I tried. It instead creates it's own stack trace of Unhanded promise rejection. Looking at it again that stack trace is being thrown elsewhere and I over complicated things. I'm going to try and find it.

Issue #377, opened 2015-05-29
Support for grammar "Gherkin" and extension "feature"


betogomez:

Glavin001:
This language? https://github.com/cucumber/cucumber/wiki/Gherkin
Could you provide a sample input and output?

betogomez:
Yes it is :)
 https://github.com/cucumber/cucumber/wiki/Gherkin

Spy-Seth:
What do you mean by "a sample input and output"? An example of un-formatted file and his fixed version?

Glavin001:
Yep. 




If the sample is similar to existing languages then we could support it easier. If it's different then we have an example to compare to our existing examples. 



—

Spy-Seth:
Here it is: https://gist.github.com/Spy-Seth/3206c3015db5eb22fd54

betogomez:
Great! looks good!

Glavin001:
Published to v0.28.9

Spy-Seth:
Excellent news! :dancers:

PR #488, opened 2015-07-30
Add Gherkin grammar support

jhansche:
This resolves Glavin001/atom-beautify#377

Uses the https://github.com/cucumber/gherkin/tree/master/js node.js
package to use the official Lexer class.

The only known issue I've seen so far is that it does not format
tables properly -- the columns are not resized to match the widest
cell in each column.

### DIFF
--- a/examples/nested-jsbeautifyrc/gherkin/expected/tables.feature
+++ b/examples/nested-jsbeautifyrc/gherkin/expected/tables.feature
@@ -1,0 +1,10 @@
+Feature: Description
+
+  Scenario Outline: This one will have examples
+    Given the <given>
+    Then the <then>
+
+    Examples:
+      | what              | then              |
+      | first value       | first then        |
+      | second what value | second then value |
--- a/examples/nested-jsbeautifyrc/gherkin/expected/test.feature
+++ b/examples/nested-jsbeautifyrc/gherkin/expected/test.feature
@@ -1,0 +1,23 @@
+Feature: The description of the feature that
+  might need to be wrapped
+  onto multiple lines
+
+  Scenario: The description of the scenario that
+    might need to be wrapped
+    onto multiple lines
+    Given I am in some state
+    When I take some action
+    # And I can add a comment
+    Then I should see some outcome
+
+  # A comment gets moved above the tag
+  @tag1
+  Scenario: This scenario has a tag
+    Then I should see some outcome
+
+  @tag1 @tag2
+  Scenario: This scenario has two tags
+    Then I should see some outcome
+
+  Scenario: This scenario does not have any tags
+    Then I should see some outcome
--- a/examples/nested-jsbeautifyrc/gherkin/original/_tables.feature
+++ b/examples/nested-jsbeautifyrc/gherkin/original/_tables.feature
@@ -1,0 +1,9 @@
+Feature: Description
+
+Scenario Outline: This one will have examples
+    Given the <given>
+    Then the <then>
+    Examples:
+    | what | then |
+    | first value | first then|
+    | second what|second then|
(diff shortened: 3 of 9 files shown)


meustice:
@Glavin001 When can we expect v0.29.0 with Gherkin support?  very excited about this update!  :+1:

Glavin001:
This is great! Thank you for this Pull Request. Only thing that is missing is tests: place `original` and `expected` files in a `gherkin` directory inside of `examples/nested-jsbeautifyrc/` directory: https://github.com/Glavin001/atom-beautify/tree/master/examples/nested-jsbeautifyrc
Thanks again!

@meustice I hope to merge this soon. I have been set back with v0.29.0 because of work and I do not have the time that I would like to review all of the Pull Requests properly and add new features under the v0.29.0 milestone. Hopefully this week goes well and I have a little bit of time this weekend to review and merge everything.
If the tests are all there, I would like to merge all of the currently open Pull Requests and then make a new release.
Thank you for your patience!

jhansche:
@Glavin001 thanks, I'll see if I can get around to writing tests for this, and I'll make the change for `debug_lexer` also. The main reason I didn't want it to run initially is that it could slow things down if it's executed when it's not needed.  Is there a Logger method to check if a certain type is enabled, before doing the rest of the work?  I.e., only do the work if log level is verbose?

Glavin001:
>  Is there a Logger method to check if a certain type is enabled, before doing the rest of the work? I.e., only do the work if log level is verbose?

Good idea. You could probably do something like:

```coffeescript
loggerLevel = atom?.config.get('atom-beautify._loggerLevel') 
if loggerLevel is 'verbose':
  # Log stuff here
```

>  I'll see if I can get around to writing tests for this, and I'll make the change for debug_lexer also. 

The tests should be very easy: an example file with some styling problems in the `original` directory, and then the correctly styled output in the `expected` directory. You can disable a test by adding `_` to the front of the filename inside of `original`. The `_` prefix in the `original` example file will cause it to be ignored / skipped. So if you could add a completely working test and then maybe a disabled test for the tables, then that would be great! Should only take a few minutes. Just something rough. Thanks!

jhansche:
@Glavin001 Are the existing test failures expected?  When I run `build-package.sh`, I get a bunch of failures:

    Finished in 4.637 seconds
    55 tests, 395 assertions, 201 failures, 0 skipped

jhansche:
@Glavin001 I ran into an issue with Atom automatically inserting a newline at the end of the file upon saving (due to the `whitespace` package).  This was causing problems because atom-beautify appears to specifically strip the trailing newlines from resolved string, as even trying to explicitly add a call to `@write_blank()`, or concatenating `\n` to the resolved string, and the test continued to fail due to the trailing newline in the expected output.

The only way to get the tests to pass was by turning off the "ensure newline" option from the whitespace package, so that I could delete the trailing newline from the original and expected files.  Then it would work.

Is that a known issue?  Anything I should have done differently to account for that?

Glavin001:
> The only way to get the tests to pass was by turning off the "ensure newline" option from the whitespace package, so that I could delete the trailing newline from the original and expected files. Then it would work.
> 
> Is that a known issue? Anything I should have done differently to account for that?

Nope, you are correct. Disabling the `whitespace` package, at least temporarily, if also what I do. Alternatively you could use the Right-Click context menu item for the file in the file tree called `Beautify File` which will not trigger the `whitespace` package.

Glavin001:
> Are the existing test failures expected? When I run build-package.sh, I get a bunch of failures:

These failures are likely because there are many beautifiers supported that use third-party CLI executables that need to be installed. For instance, to run the tests for beautifying PHP code Atom Beautify requires PHP-CS-Fixer, which I assume you may not have installed. As long as your tests for Gherkin pass, then I can review this and still merge it. Thanks.

Glavin001:
Published to v0.28.9

PR #497, opened 2015-08-06
bug/nullgrammar

kelvix:
#487 and #482 

I should try sleeping and coming back to things sometimes. I think this one is better. You still get the warning pop up rightfully saying there is no support for Null Grammar, but you still get debug output.


Glavin001:
Thank you! This is passing on Travis CI so I will merge this.

Glavin001:
Published to v0.28.10

PR #501, opened 2015-08-10
Properly dispose of event subscriptions

mnquintana:
This PR updates atom-beautify so that it properly disposes of all its subscriptions by adding [a standard `deactivate` for clean-up](https://atom.io/docs/v1.0.4/hacking-atom-package-word-count#source-code). This should prevent memory leaks when updating the package, allow it to report updates successfully (see https://github.com/atom/settings-view/issues/532), and avoid potential issues with commands not working immediately after an update.

### DIFF
--- a/src/beautify.coffee
+++ b/src/beautify.coffee
@@ -4,6 +4,7 @@ pkg = require('../package.json')
 
 # Dependencies
 plugin = module.exports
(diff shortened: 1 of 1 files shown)


Glavin001:
Thank you!

Glavin001:
Published to v0.28.11

Issue #506, opened 2015-08-12
Beautify not yet saved files


philiplb:
Hi,

sometimes, I just want to beautify some code in my clipboard. For this, I'd like to throw it in an unsaved, new file in Atom and just hit "Beautify". But currently, I get this message: "Atom Beautify could not find a supported beautifier for this file".
Of course, the plugin doesn't have a clue what to do with this text.
It would be very neat, if Beautify could ask what beautifier to use in this case. The user could select then from a list.

mkieblesz:
I think we might have problem with recognizing file type here, but I have the same need.

+1

Glavin001:
This is a two part issue: 
- Atom should be attempting to recognize the grammar from the file without using its extension
- Atom Beautify should popup asking the user which beautifier they want to use, if Atom has been unable to detect the grammar

>  For this, I'd like to throw it in an unsaved, new file in Atom and just hit "Beautify". 

When I do this, I simply change the Atom text editor's grammar to my desired grammar. Go to Atom menu -> Edit -> Select grammar. You can treat that file as an already built-in way to do what you want:
> It would be very neat, if Beautify could ask what beautifier to use in this case. The user could select then from a list.

Hope that helps, it works quickly and easily for me when I do exactly what you are doing.

mkieblesz:
wow that was easy. Thanks @Glavin001

Glavin001:
You're welcome :smile:.

philiplb:
Nice, good hint. :)

PR #516, opened 2015-08-18
Can configure a fallback for unsupported language

seebigs:
I often like to paste content into a new, unsaved window and beautify it. Unfortunately, this causes Atom Beautify to throw an error. I suggest an improvement that allows users to configure a fallback for any unsupported language type.

https://gist.github.com/seebigs/ff0fc0c71d6d9ea8a9e2

### DIFF
--- a/src/beautifiers/index.coffee
+++ b/src/beautifiers/index.coffee
@@ -325,6 +325,12 @@ module.exports = class Beautifiers extends EventEmitter
         languages = @languages.getLanguages({grammar, extension: fileExtension})
         logger.verbose(languages, grammar, fileExtension)
 
+        # use fallback if configured
+        unsupportedLanguageFallback = atom.config.get("atom-beautify.unsupportedLanguageFallback")
+        if languages.length < 1 && unsupportedLanguageFallback.charAt(0) != '-'
+          languages = [require('../languages/' + unsupportedLanguageFallback)]
+          logger.verbose('Unsupported Language, fallback to:', unsupportedLanguageFallback)
+
         # Check if unsupported language
         if languages.length < 1
           unsupportedGrammar = true
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -1,3 +1,9 @@
+Languages = require('./languages/')
+
+langFallbackDefault = '- none (do not format) -'
+langFallbacks = new Languages().languageNames
+langFallbacks.unshift(langFallbackDefault)
+
 module.exports = {
   analytics :
     type : 'boolean'
@@ -28,4 +34,9 @@ module.exports = {
     type : 'boolean'
     default : false
     description : "Do not show any/all errors when they occur"
+  unsupportedLanguageFallback :
+    type : 'string'
+    default : langFallbackDefault
+    description : "Format any Unsupported Languages as the language you specify here"
+    enum : langFallbacks
 }
--- a/package.json
+++ b/package.json
@@ -50,6 +50,10 @@
     {
       "name": "Clemens Damke",
       "url": "https://github.com/Cortys"
+    },
+    {
+      "name": "Chris Bigelow",
+      "url": "https://github.com/seebigs"
     }
   ],
   "engines": {


Glavin001:
> I often like to paste content into a new, unsaved window and beautify it. 

This use case sounds similar to what was described on another issue. I recommend reading over https://github.com/Glavin001/atom-beautify/issues/506#issuecomment-130362853

> I suggest an improvement that allows users to configure a fallback for any unsupported language type.

The language likely is not unsupported, and actually has simply not been detected by Atom, yet. Reading https://github.com/Glavin001/atom-beautify/issues/506#issuecomment-130362853 should help resolve that issue.

seebigs:
Sorry, I searched issues for similar cases but didn't find #506. Thanks for pointing that out.

I still think this would be an improvement because it saves keystrokes, and honestly, it's just poor UX to throw an error message in such a scenario. Perhaps prompting the user to choose a grammar ***is*** better (and optionally check a box to always use the selected grammar when language is unsupported). Or maybe you could open Atom's Grammar Selector automatically, then proceed with beautification after a grammar has been chosen for that file.

I'm only being so picky about this because my favorite Sublime plugin has no problem with this use case:
https://packagecontrol.io/packages/HTML-CSS-JS%20Prettify
This plugin just beautifies the unsaved file anyway without even prompting which grammar to use, and this has become a part of my daily workflow. Your error message blocks this and makes me want to find alternatives.

Glavin001:
First I want to address why `HTML-CSS-JS Prettify` plugin for Sublime works.

So this is how `HTML-CSS-JS Prettify` plugin for Sublime does it:
1. Code is read from your Sublime edit and sent to `scripts/run.js`. See https://github.com/victorporof/Sublime-HTMLPrettify/blob/master/HTMLPrettify.py#L86-L87
2. In `scripts/run.js` the code and filePath is extracted from the command-line arguments passed to it. See https://github.com/victorporof/Sublime-HTMLPrettify/blob/master/scripts/run.js#L18-L19
3. It attempts to determine the type of the code, checking `isCSS`, isHTML` and `isJS`. See https://github.com/victorporof/Sublime-HTMLPrettify/blob/master/scripts/run.js#L63-L71
4. If there is an extension, then it uses that. This is not our case: we have no extension for this unsaved file. 
  - `isCSS` will not work if there is no extension, see https://github.com/victorporof/Sublime-HTMLPrettify/blob/master/scripts/run.js#L132-L135
  - `isHTML` will look for `<` character and return true if it exists
  - `isJS` will look for `<` character and return false if it exists

Two problems:
- CSS is not supported. Try it.
- Given the above logic, I theorize that if you were to create a new, unsaved JavaScript file with the following contents it would not actually beautify as JavaScript and maybe incorrectly interpreted as HTML.

```javascript
function example() {
var broken = "<";
}
```


Regardless, JS Beautify's website does correctly handle this case with their check: https://github.com/beautify-web/js-beautify/blob/master/index.html#L287-L294

```javascript
function looks_like_html(source) {
            // <foo> - looks like html
            // <!--\nalert('foo!');\n--> - doesn't look like html
            var trimmed = source.replace(/^[ \t\n\r]+/, '');
            var comment_mark = '<' + '!-' + '-';
            return (trimmed && (trimmed.substring(0, 1) === '<' && trimmed.substring(0, 4) !== comment_mark));
        }
```

Try at http://jsbeautifier.org/

---

Atom Beautify supports a very wide variety of languages and as such having a *single* catch-all fallback is simply unacceptable for an advanced user. This is why I always recommend choosing your grammar, in Atom, as that is how other Atom packages will also be detect the grammar/language of your file.

These other packages can perform very primitive checks to attempt to detect the language, however they also require maintaining and bug fixes, and it is already a *very* easy task, for a human, to determine.

That being said, for non-advanced users, who are only beautifying a very small subset of languages with Atom Beautify, it would be nice for them to always use a fallback beautifier. Such as a user who is a website designer and only uses HTML and CSS, they may want to simply say that HTML language is their fallback and let Atom Beautify perform that beautification. In the case where it is not HTML, then it is, again, on the onus of the user to pre-select the proper grammar. The advantage being that for the most part, Atom Beautify would attempt to beautiful before causing trouble -- such as displaying an error saying the beautification failed because the beautifier could not process the language, as it was not actually HTML.
I think something like this should satisfy you and other users. 

In its current state, this Pull Request is not something I would merge. I will comment in your Pull Request commits with some recommendations of improvements to make this merge-able.

Thank you for your contributions and feedback.

src/beautifiers/index.coffee
```diff
@@ -325,6 +325,12 @@ module.exports = class Beautifiers extends EventEmitter
         languages = @languages.getLanguages({grammar, extension: fileExtension})
         logger.verbose(languages, grammar, fileExtension)
 
+        # use fallback if configured
+        unsupportedLanguageFallback = atom.config.get("atom-beautify.unsupportedLanguageFallback")
+        if languages.length < 1 && unsupportedLanguageFallback.charAt(0) != '-'
+          languages = [require('../languages/' + unsupportedLanguageFallback)]
```
Glavin001: This should be changed to something like:

```coffee
languages = @languages.getLanguages({name: unsupportedLanguageFallback})
```

See https://github.com/Glavin001/atom-beautify/blob/master/src/languages/index.coffee#L82 for details.

seebigs: @Glavin001 
I wanted to do that... but the problem is that I need an enum of all of the supported languages to populate the config dropdown (https://github.com/Glavin001/atom-beautify/blob/master/src/config.coffee), so I use `languageNames` from https://github.com/Glavin001/atom-beautify/blob/master/src/languages/index.coffee#L14

Then, for example, one of the languageNames that I might pick is `'javascript'`

`@languages.getLanguages` wants `name` to be something case sensitive like `'JavaScript'` which is not available in any one enum, but only from each individual language module
https://github.com/Glavin001/atom-beautify/blob/master/src/languages/javascript.coffee#L11

Do you have a suggestion for how to get the enum differently?

src/config.coffee
```diff
@@ -1,3 +1,9 @@
+Languages = require('./languages/')
+
+langFallbackDefault = '- none (do not format) -'
+langFallbacks = new Languages().languageNames
```
Glavin001: Fortunately, by creating a new `Languages` instance, we have access to all of the loaded `languages` and their `name` property.

Instead, let's try something like:

```coffee
# Add lodash to the requirements at the top, since it is not already
_ = require('lodash')
# Map from the Languages to each language.name
langFallbacks = _.map((new Languages()).languages, (language) => language.name)
```

It essentially is the opposite of how the `getLanguages` works, see https://github.com/Glavin001/atom-beautify/blob/master/src/languages/index.coffee#L85 , since it gets the language `name`s the same way.

src/beautifiers/index.coffee
Glavin001: I just added a comment: https://github.com/Glavin001/atom-beautify/pull/516/files#r37490852 . Let me know if that helps.

src/config.coffee
```diff
@@ -28,4 +35,9 @@ module.exports = {
     type : 'boolean'
     default : false
     description : "Do not show any/all errors when they occur"
+  unsupportedLanguageFallback :
+    type : 'string'
```
Glavin001: Add `title` field.

PR #540, opened 2015-09-08
Add extra_liners option from js-beautify in Settings View

zedentox:
Hi !

This PR add a "HTML - Extra liners" field in Settings View. 
See extra_liners option for HTML docs here : https://github.com/beautify-web/js-beautify#css--html

This option is already supported by atom-beautify but you need to edit your config file manually.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -916,6 +916,36 @@ End output with newline (Supported by JS Beautify)
 }
 ```
 
+####  [HTML - Extra liners](#html---extra-liners) 
+
+**Namespace**: `html`
+
+**Key**: `extra_liners`
+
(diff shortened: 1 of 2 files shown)


Glavin001:
Great! Could you run `coffee docs/` to generate the documentation for this and then commit those changes as well. Thank you!

zedentox:
Done. 

I wanted to update the changelog but I have a doubt about which version it concerns. v0.29.0 ?

Glavin001:
Looks great! I'll fix the failing tests, they are not a problem with this pull request.

Glavin001:
Thank you for contributing!

PR #546, opened 2015-09-12
Add two beautifiers: clang-format and yapf

ProgramFan:
clang-format is a semantics-aware C/C++/Objective-C beautifier from clang compiler tool. It produces correct code even for very complicated sources. So this can be an alternative to uncrusify for C/C++/Obj-C.

yapf (Yet Another Python Formatter) is a python formatter using the same format algorithm as clang-format. It's said to produce better looking pep8-conforming code than autopep8. So someone might want it.

This pull request implements two beautifiers for the above code formatters. I have tested it with my own code base and it works fine.

### DIFF
--- a/src/beautifiers/clang-format.coffee
+++ b/src/beautifiers/clang-format.coffee
@@ -65,8 +65,8 @@ module.exports = class ClangFormat extends Beautifier
         @dumpToFile(dumpFile, text)
         ["--style=file"]
         ], help: {
-        link: "https://clang.llvm.org/docs/ClangFormat.html"
-        }).finally( =>
+          link: "https://clang.llvm.org/docs/ClangFormat.html"
+        }).finally( ->
           fs.unlink(dumpFile)
         )
     )
--- a/src/beautifiers/clang-format.coffee
+++ b/src/beautifiers/clang-format.coffee
@@ -36,8 +36,20 @@ module.exports = class ClangFormat extends Beautifier
     )
 
   beautify: (text, language, options) ->
+    # NOTE: One may wonder why this code goes a long way to construct a file
+    # path and dump content using a custom `dumpToFile`. Wouldn't it be easier
+    # to use `@tempFile` instead? The reason here is to work around the
+    # clang-format config file locating mechanism. As indicated in the manual,
+    # clang-format (with `--style file`) tries to locate a `.clang-format`
+    # config file in directory and parent directories of the input file,
+    # and retreat to default style if not found. Projects often makes use of
+    # this rule to define their own style in its top directory. Users often
+    # put a `.clang-format` in their $HOME to define his/her style. To honor
+    # this rule, we HAVE TO generate the temp file in THE SAME directory as
+    # the editing file. However, this mechanism is not directly supported by
+    # atom-beautify at the moment. So we introduce lots of code here.
     return new @Promise((resolve, reject) ->
-      editor = atom.workspace.getActiveTextEditor()
+      editor = atom?.workspace?.getActiveTextEditor()
       if editor?
         fullPath = editor.getPath()
         currDir = path.dirname(fullPath)
@@ -54,7 +66,7 @@ module.exports = class ClangFormat extends Beautifier
         ["--style=file"]
         ], help: {
         link: "https://clang.llvm.org/docs/ClangFormat.html"
-      }).then( =>
-        fs.unlinkSync(dumpFile)
-      )
+        }).finally( =>
+          fs.unlink(dumpFile)
+        )
     )
--- a/src/beautifiers/clang-format.coffee
+++ b/src/beautifiers/clang-format.coffee
@@ -12,9 +12,9 @@ module.exports = class ClangFormat extends Beautifier
   name: "clang-format"
 
   options: {
-    "C++": true
-    "C": true
-    "Objective-C": true
+    "C++": false
+    "C": false
+    "Objective-C": false
   }
 
   ###
--- a/src/beautifiers/clang-format.coffee
+++ b/src/beautifiers/clang-format.coffee
@@ -1,0 +1,60 @@
+###
+Requires clang-format (https://clang.llvm.org)
+###
+
+"use strict"
+Beautifier = require('./beautifier')
+path = require('path')
+fs = require('fs')
+
+module.exports = class ClangFormat extends Beautifier
+
+  name: "clang-format"
+
+  options: {
+    "C++": true
+    "C": true
+    "Objective-C": true
+  }
+
+  ###
+    Dump contents to a given file
+  ###
+  dumpToFile: (name = "atom-beautify-dump", contents = "") ->
+    return new Promise((resolve, reject) =>
+      fs.open(name, "w", (err, fd) =>
+        @debug('dumpToFile', name, err, fd)
+        return reject(err) if err
+        fs.write(fd, contents, (err) ->
+          return reject(err) if err
+          fs.close(fd, (err) ->
+            return reject(err) if err
+            resolve(name)
+          )
+        )
+      )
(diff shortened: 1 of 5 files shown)


src/beautifiers/clang-format.coffee
```diff
@@ -0,0 +1,60 @@
...
+fs = require('fs')
+
+module.exports = class ClangFormat extends Beautifier
+
+  name: "clang-format"
+
+  options: {
+    "C++": true
+    "C": true
+    "Objective-C": true
```
Glavin001: It does not look like any of the `options` are used, so change it from `true` to `false` to indicate that while the language is supported it does not support any of the options directly.

src/beautifiers/clang-format.coffee
```diff
@@ -0,0 +1,60 @@
...
+        currFile = path.basename(fullPath)
+        dumpFile = path.join(currDir, ".atom-beautify.#{currFile}")
+        resolve dumpFile
+      else
+        reject(new Error("No active editor found!"))
+    )
+    .then((dumpFile) =>
+      # console.log("clang-format", dumpFile)
+      @run("clang-format", [
+        @dumpToFile(dumpFile, text)
```
Glavin001: You should use `tempFile("input", text)`, like you did below in the yapf beautifier.

src/beautifiers/clang-format.coffee
```diff
@@ -0,0 +1,60 @@
...
+            return reject(err) if err
+            resolve(name)
+          )
+        )
+      )
+    )
+
+  beautify: (text, language, options) ->
+    return new @Promise((resolve, reject) ->
+      editor = atom.workspace.getActiveTextEditor()
```
Glavin001: Should remove all of this code requesting the editor, editor path, and other directory operations. Use `tempFile('input', text)` which generates a generic path and temporary file that gets automatically deleted.

src/beautifiers/clang-format.coffee
ProgramFan: Thanks for pointing it out. Since clang-format uses its own configuration file, it does not make too much sense to accept options. And projects using clang-format usually provides their own configuration file. I will change it in a new commit.

src/beautifiers/clang-format.coffee
ProgramFan: Thanks for your careful review. :+1:  But since clang-format uses a directory based approach to find the configuration file, we need to provide it with correct path. 

To be specific, to format file `/aa/bb/cc/dd/ee/ff.cpp`, clang-format will find in directories `[/aa/bb/cc/dd/ee, /aa/bb/cc/dd, /aa/bb/cc, /aa/bb, /aa, /]` in the apperance order for a file `.clang-format`. Project usually put this file in the project root and users would like to put a global `.clang-format` in their home directory. If we use tempFile, the file is usually created in `/tmp/xxx`, so no `.clang-format` can be found in the way, which causes clang-format uses default format style, which is only good for llvm (whose format is the default of clang-format).

To workaround the above problem, I request the editor path and constructed a temp file in the same directory as the original file, and delete it after use. This is why the code deliberately creates a 'dumpToFile' function.

If there are better approaches, I would like very much to explore it.

ProgramFan: If the current solution is acceptable, I would add a comment in the code so it can be less confusing.

src/beautifiers/clang-format.coffee
```diff
@@ -0,0 +1,60 @@
...
+    )
+    .then((dumpFile) =>
+      # console.log("clang-format", dumpFile)
+      @run("clang-format", [
+        @dumpToFile(dumpFile, text)
+        ["--style=file"]
+        ], help: {
+        link: "https://clang.llvm.org/docs/ClangFormat.html"
+      }).then( =>
+        fs.unlinkSync(dumpFile)
```
Glavin001: Could you make this not the `sync` function call? Wouldn't want to lock up Atom is the hard drive is slow. This will allow other things to be processed if the hard drive is taking time.

src/beautifiers/clang-format.coffee
Glavin001: I think the current solution is acceptable. I will have to consider giving the Beautifier class access to the file path and not only its contents.

You could change:

```coffee
```

to 

```coffee
editor = atom?.workspace?.getActiveTextEditor()
```

Such that it does not error if it does not run within Atom, and `editor` will then be `null`.

This is because beautification testing may run from outside of Atom and I want to make sure it works (does not crash) and throws errors if it does require Atom (which this does).

Thanks!

Glavin001:
I am fixing the tests now. Please re-merge the latest changes, fix the above comments, and I can review your Pull request and merge! :smiley:  Thank you!

src/beautifiers/clang-format.coffee
ProgramFan: Thanks, Glavin. I have made necessary changes and they are waiting for your review. However, a strange issue arises in the process. The beautifier does nothing to the buffer with atom-1.0.15, while it works correctly under atom-1.0.9. It can be fixed by changing the last ".then" to ".finally". I am new to the javascript world and do not quite understand the Promise mechanisms. Would you please suggest a better or more idiomatic approach to delete the temp file after the @run call?

Glavin001: `then` will be called if all of the Promises passed, successfully. `finally` will be called regardless of if it was successful (`then`) or an error (`catch`). So it sounds like an error was occuring and it was not running the `then` callback. Makes good sense to put it under the `finally` callback.

src/beautifiers/clang-format.coffee
```diff
@@ -0,0 +1,72 @@
...
+        currDir = path.dirname(fullPath)
+        currFile = path.basename(fullPath)
+        dumpFile = path.join(currDir, ".atom-beautify.#{currFile}")
+        resolve dumpFile
+      else
+        reject(new Error("No active editor found!"))
+    )
+    .then((dumpFile) =>
+      # console.log("clang-format", dumpFile)
+      @run("clang-format", [
```
Glavin001: I think this needs to be `return`ed.

```
return @run`
```

therefore its results (success or error) are passed up to the promise above, which is already returned.

Glavin001: Actually maybe CoffeeScript does correctly return this one already.

I was just thinking that one of your comments sounded like an error was occurring (`then` not being called) and I thought maybe this could be why.

Were you able to have a test that did return an error and properly notify you?

ProgramFan: I have tried the following tests:
A. Null workspace. By running `beautify buffer` on the `settings` tab, an error is notified by `atom-beautify`, with the content `no workspace found`.
B. Readonly directory. By changing the mode of the containing directory to 'u-w', an error with `EACCS` is notified. This works with and without `return @run...`.

I doubt that chrome changes its implementation of promises, since atom upgrades its chrome to 43? in at least 1.0.15. The `then` works correctly (the temp file is deleted) but the stdout of `@run` is not passed out correctly. Although I am still curious why it works, the `finally` solves the problem.

Glavin001:
I think this is almost ready to merge! Looking at the failing tests, it appears that linting has failed. You can run `npm run lint` to try for yourself. Once linting passes and the rest of the tests pass (which I think they will) I can merge! Great work! Thank you for contributing!

ProgramFan:
The lint errors are fixed. It seems that the appveyor build failed because it's unable to install language-svg and it shall not be caused by the commit.

Glavin001:
Thank you!

ProgramFan:
It's a pleasure to contribute. :)

Glavin001:
Published to v0.28.15

PR #568, opened 2015-09-24
Support dynamically reloading custom file .csscomb.json file changes

DanielJoyce:
Instead of using require which caches the result, check if file exists and then use it if it does, otherwise fall back to bundled config.

### DIFF
--- a/src/beautifiers/csscomb.coffee
+++ b/src/beautifiers/csscomb.coffee
@@ -1,5 +1,6 @@
 "use strict"
 Beautifier = require('./beautifier')
+fs = require('fs')
 
 module.exports = class JSBeautify extends Beautifier
   name: "CSScomb"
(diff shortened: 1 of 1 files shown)


DanielJoyce:
Fails, but due to formatting issues involving other modules.

src/beautifiers/csscomb.coffee
```diff
@@ -23,7 +24,12 @@ module.exports = class JSBeautify extends Beautifier
       try
         # Load from Project's .csscomb.json
         projectConfigPath = atom.project.getDirectories()?[0]?.resolve('.csscomb.json')
-        config = require(projectConfigPath) # Will throw error if does not exist
+        #config = require(projectConfigPath) # Will throw error if does not exist
+
+        stats = fs.statSync(projectConfigPath)
```
Glavin001: Please do not use the `sync` form of this function.

If you could re-write this using the asynchronous function then I think this is good to review and merge.

Glavin001:
Thanks!

Issue #575, opened 2015-09-27
JSP support


duvanjamid:
it would be quite useful !!

prettydiff:
Pretty Diff can already support this.  One of us just needs to create a pull request exactly like: https://github.com/Glavin001/atom-beautify/issues/559

Glavin001:
An example Pull request being #560

kinkwilde:
Sorry to bring this one up again - JSP support is provided in XML langauge support, but only the extension .jsp is support - can we get .jspf supported as well? At the moment I have to change the language from javascript pages to xml in atom before being able to beautify.

PR #586, opened 2015-10-03
JSP support via beautifying as XML in Pretty Diff

prettydiff:
Issue #575


Glavin001:
Thanks!

jaibatrik:
@Glavin001 Since this is already merged, when will it make it to the distributed plugin? Looks like beautifying JSP still does not work with atom-beautify.

Glavin001:
This was merged in 2015. This should be available now. If it is not available then I recommend that you create an Issue and be sure to include your `debug.md` Gist that we can use to diagnose the problem.

PR #590, opened 2015-10-05
Issue 589

prettydiff:


Glavin001:
Could you explain why this is necessary? Sounds like [Pretty Diff's `package.json` has an incorrect `main`](https://github.com/prettydiff/prettydiff/blob/master/package.json#L25) such that `require('prettydiff')` does not work properly.

prettydiff:
Ahhhhhhh, that could be it.  I did not think to look at that.

prettydiff:
Reverted changes to package.json main in v1.14.2.  This appears to have fixed the problem.

Issue #591, opened 2015-10-06
Method Chaining issue with prettydiff (jsx)


hayatae:
I'm not sure if behavior recently changed in prettydiff or atom-beautify or both, but when I beautify React (jsx) files, all of my chained methods are now broken up into individual lines instead of remaining on a single line. This never used to be a problem.

According to the atom-beautify setting for "Break chained methods" in JavaScript, this setting is only supported by JS Beautify. I don't see why this can't be supported by Pretty Diff as well since Pretty Diff has a setting for "methodchain" (http://prettydiff.com/documentation.xhtml). Granted its boolean value would be the opposite. As atom-beautify currently is, there is no way to adjust the method chaining setting that Pretty Diff uses, which means it always defaults to breaking chained methods.

For a workaround, I tried creating a .prettydiffrc file in my project root with the following settings:
{
  "methodchain": true
}

Unfortunately that didn't do the trick. I'm not sure if the file was ignored or if atom-beautify is overriding the settings.

Not sure if this is an issue for @prettydiff or atom-beautify or both.

Are there any workarounds or solutions for this?

prettydiff:
This is due to a recent Pretty Diff change.  It is a very new method to Pretty Diff as I published it a little more than 2 weeks ago: https://github.com/prettydiff/prettydiff/issues/161

Look at this issue to track the status of Atom-Beautify support for new Pretty Diff options: https://github.com/Glavin001/atom-beautify/issues/524

> Granted its boolean value would be the opposite.

Before I published this feature/option to break method chains I did not support this in any way.  Once I added support I made the current behavior of breaking method chains the default behavior.  I provided the option only as a means to override this new default behavior.  Since the option is merely an override that returns method lists into a chain on a single line I named the option `methodchain` and its default value is `false` which corresponds to the default behavior of breaking chains onto individual lines.

I honestly didn't look to see if js-beautify had such an option, what its default behavior is, or how its naming convention/values work.

> I tried creating a .prettydiffrc file in my project root

The Atom-Beautify project does not use the `.prettydiffrc` file.  This project has its own internal option system that it has to use to bridge support from various formatting applications into a single API.  That and also the `.prettydiffrc` file is also very new at right about a month old.

What you can do for a work around is add this comment to your code:

    /*prettydiff.com methodchain:true*/

The order of precedence for options in Pretty Diff is the following from greatest to least:

1. The styleguide option (this only applies to the jspretty library)
2. The prettydiff comment
[...]

Issue #524, opened 2015-08-23
New options added to Pretty Diff


prettydiff:
Just a quick FYI. Over the past month or so a couple new options have been added to the Pretty Diff tool:

* [commline](http://prettydiff.com/documentation.xhtml#commline) - https://github.com/prettydiff/prettydiff/issues/139
* [noleadzero](http://prettydiff.com/documentation.xhtml#noleadzero) - https://github.com/prettydiff/prettydiff/issues/146
* [spaceclose](http://prettydiff.com/documentation.xhtml#spaceclose) - https://github.com/prettydiff/prettydiff/issues/139
* [tagsort](http://prettydiff.com/documentation.xhtml#tagsort) - https://github.com/prettydiff/prettydiff/issues/89

Glavin001:
Awesome, thanks @prettydiff !

For anyone interested in making a Pull Request and want to know how:
- Applicable Pretty Diff beautifier code: https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/prettydiff.coffee#L7-L29
- Make sure the allowed languages ( from https://github.com/Glavin001/atom-beautify/tree/master/src/languages ) have those options available.

Try to keep the naming standardized. For example, Atom Beautify should have one way to name an option and it should use beautifier mapps to convert naming of options to JS Beautify and Pretty Diff conventions.

prettydiff:
Just added a new option `endcomma` in resolution to https://github.com/Glavin001/atom-beautify/issues/534

I am not sure if this will be helpful for you, but I just added support for a `.prettydiffrc` file.  The way it works is that the node-local.js file will first look for this file in the directory currently active in the command prompt and if not available then checks for this file in the application's directory.  The file is first checked for JSON format and if not JSON format checks to see if the file is a JavaScript application.  The file is not evaluated until after options supplied from the command line are evaluated. The JavaScript application format is available to allow optional complexity so that users can define options conditional upon options supplied from the command line, such as defining options per language, mode, or a combination of options.  Here is an example of the file in JavaScript format: http://prettydiff.com/.prettydiffrc

Glavin001:
> I am not sure if this will be helpful for you, but I just added support for a .prettydiffrc file.

I think this may be very helpful for users who want to use new Pretty Diff features that Atom Beautify has not yet added support for. Thanks!

Glavin001:
[...]

PR #594, opened 2015-10-08
See #524. Added support for methodchain setting in prettydiff.

hayatae:
This is to help address #524 and #591. It looks like the docs were a bit out of date when I ran coffee /docs, so there's some extra `clang-format` support noted in there that's unrelated to this change.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -1254,11 +1254,11 @@ Add a space before an anonymous function's parens, ie. function () (Supported by
 
 **Type**: `boolean`
 
-**Supported Beautifiers**:  [`JS Beautify`](#js-beautify) 
+**Supported Beautifiers**:  [`JS Beautify`](#js-beautify)  [`Pretty Diff`](#pretty-diff) 
 
 **Description**:
 
-Break chained method calls across subsequent lines (Supported by JS Beautify)
+Break chained method calls across subsequent lines (Supported by JS Beautify, Pretty Diff)
 
 **Example `.jsbeautifyrc` Configuration**
 
@@ -2007,7 +2007,7 @@ Disable C Beautification
 
 **Type**: `string`
 
-**Enum**:  `Uncrustify` 
+**Enum**:  `Uncrustify`  `clang-format` 
 
 **Description**:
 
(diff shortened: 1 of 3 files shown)


hayatae:
Whoops, I see that a unit test needs to be updated for the new setting. I'll get that fixed and do another request.

hayatae:
Oh great. Now the CI tests failed because they couldn't download atom from atom.io.  Can someone with permission re-run the CI tests?

Glavin001:
I restarted both CIs.

hayatae:
Thanks! Not sure why appveyor failed when it reported 0 failures though:
Finished in 16.949 seconds
54 tests, 194 assertions, 0 failures, 0 skipped
Tests failed
Command exited with code 1

On a related question, how do you run tests locally? I tried running `apm test` in the package root, but I get a whole bunch of errors that look like:
[57512:1009/010335:INFO:CONSOLE(410)] "Could not resolve 'language-svg' to a package path", source: /usr/share/atom/resources/app.asar/src/package-manager.js (410)

Glavin001:
Running tests locally takes a little work. You have to install everything, including [specific Atom packages (such as `language-svg` as you noted above)](https://github.com/Glavin001/atom-beautify/blob/master/.travis.yml#L22), and [third-party beautifiers with Operating System specific installation instructions](https://github.com/Glavin001/atom-beautify/blob/master/.travis.yml#L31-L55), and then [run all of the tests](https://github.com/Glavin001/atom-beautify/blob/master/build-package.sh#L42) for every language and their default supported beautifier. It takes Travis CI under 5 minutes to do all of that.
You can take a look at https://github.com/Glavin001/atom-beautify/blob/master/.travis.yml#L19-L55 and https://github.com/Glavin001/atom-beautify/blob/master/build-package.sh to understand how Travis CI does the tests.

Glavin001:
Thank you for contributing!

zachguo:
Is this in latest release (0.28.16)? I didn't check `break_chained_methods`, but it's still breaking method chains in JSX.

hayatae:
Unfortunately this didn't make it into 28.16.  We'll have to wait for the next release. It looks like there's some other major issues going on with prettydiff right now which also need to be fixed though. I still can't use atom-beautify on my files right now. :<

Glavin001:
Published to v0.28.17

Issue #209, opened 2015-02-09
CSSComb support


alem0lars:
The current CSS beautifier is not really flexible.

I think [CSScomb](http://csscomb.com) would be much better since it's a more [configurable](http://csscomb.com/config) and flexible beautifier.

We can take a look at our people implemented it in atom, here there are some examples:
- `atom-css-comb`
- `css-comb`

Thanks

prettydiff:
@alem0lars the links provided for CSScomb resolve to 404 pages. Is there some other location for CSScomb?

alem0lars:
Yes: [Github page](https://github.com/csscomb/csscomb.js)

danielbayley:
+1

Glavin001:
Links to existing Atom packages:
- [atom-css-comb](https://atom.io/packages/atom-css-comb)
- [css-comb](https://atom.io/packages/css-comb)
- [atom-csscomb](https://atom.io/packages/atom-csscomb)

Glavin001:
CSScomb will be included in v0.26.0

To switch from `JS Beautify` (CSS)` or `Pretty Diff` (SCSS/Sass/LESS) to using `CSScomb`, you will go into Atom Beautify package settings and find the option for `Language Config - <Your Desired Language, CSS, etc> - Default Beautifier` and select `CSScomb`.

CSScomb looks for a `.csscomb.json` file in your Atom's first project directory. If it does not find one, or if importing it fails, it will fallback to `csscomb` predefined configuration.

Glavin001:
Published to v0.26.0

alem0lars:
Thanks very much :smile:

DanielJoyce:
Buggy, doesn't seem to reload .csscomb.json sometimes.

wyudong:
It must restart Atom to let custom `configPath` to take effect.

Glavin001:
@DanielJoyce : The current implementation caches `.csscomb.json` and would require reloading Atom to clear cache: https://github.com/Glavin001/atom-beautify/pull/623/files#diff-da1ec4e677f98ecd91454c8d30ff68eaR33
Using `.csscomb.cson` does not appear to cache and should work without reloading: https://github.com/Glavin001/atom-beautify/pull/623/files#diff-da1ec4e677f98ecd91454c8d30ff68eaR31
Please submit a Pull Request implementing anything fixes you might have.

robsonenke:
[...]

PR #623, opened 2015-10-28
Improve CSScomb support

danielbayley:
Add setting for custom config file path and support for CSON in addition to the standard JSON config.

Signed-off-by: Daniel Bayley <daniel.bayley@me.com>

### DIFF
--- a/src/beautifiers/csscomb.coffee
+++ b/src/beautifiers/csscomb.coffee
@@ -6,10 +6,13 @@ module.exports = class JSBeautify extends Beautifier
 
   options: {
     # TODO: Add support for options
-    CSS: false
-    LESS: false
-    Sass: false
-    SCSS: false
+    _:
+      configPath: true
+      predefinedConfig: true
+    CSS: true
+    LESS: true
+    Sass: true
+    SCSS: true
   }
 
   beautify: (text, language, options) ->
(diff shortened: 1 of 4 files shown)


danielbayley:
Following on from #209…

danielbayley:
@Glavin001 Not sure why this is failing? I don't think it has anything to do with the code I added…

src/beautifiers/csscomb.coffee
```diff
@@ -6,10 +6,10 @@ module.exports = class JSBeautify extends Beautifier
 
   options: {
     # TODO: Add support for options
-    CSS: false
-    LESS: false
-    Sass: false
-    SCSS: false
+    CSS: true
```
Glavin001: Change:

```
    CSS: true
    LESS: true
    Sass: true
    SCSS: true
```

to:

```
    _:
      configPath: true
      predefinedConfig: true
    CSS: true
    LESS: true
    Sass: true
    SCSS: true
```

Similar to https://github.com/danielbayley/atom-beautify/blob/master/src/beautifiers/prettydiff.coffee#L6

This allows only the options that are supported (`configPath` and `predefinedConfig`) and puts that for the default (`_`).

src/beautifiers/csscomb.coffee
```diff
@@ -18,16 +18,25 @@ module.exports = class JSBeautify extends Beautifier
 
       # Require
       Comb = require('csscomb')
+      expandHomeDir = require('expand-home-dir')
+      CSON = require('season')
 
+      readConfig = (configFile) ->
+        try
+          CSON.readFileSync(configFile)
```
Glavin001: Does this also work for `.json` files? Has this been tested?

Glavin001:
I usually ignore passing AppVeyor since Windows can be a PITA.

It is currently passing Travis CI, which is what I require before merging Pull Requests, so this is good :+1:.

After you fix the above recommended changes, please run `coffee docs/` as well which will generate the documentation with the new options you have added.

This is great. Good work so far! Will be able to merge this soon. Thanks.

danielbayley:
`CSON.readFileSync` seems to actually read JSON too; so Iv'e simplified the logic to take that into account…

> since Windows can be a PITA.

*Can* be?!

src/beautifiers/csscomb.coffee
Glavin001: This is the last thing that needs fixing. Right now it looks like all options are available, when it should be only `configPath` and `predefinedConfig` option. After that, re-run `coffee docs/` to generate the correct documentation with only the allowed options. Thanks!

Glavin001:
> CSON.readFileSync seems to actually read JSON too; so Iv'e simplified the logic to take that into account…

Great!

danielbayley:
All sorted now as far as I can tell. The CI build is failing but since I'm getting deprecation warnings that include this package (since this last update to Atom itself) I'm guessing it's something to do with that…

Glavin001:
It's actually failing because of the Titanium Style Sheets tests.
I'm going to merge this and take a look on my own.

Glavin001:
Thank you!

Glavin001:
Published to v0.28.17

Issue #628, opened 2015-10-29
Support stylish-haskell


bobjflong:
https://github.com/jaspervdj/stylish-haskell

Working on a PR for this

PR #629, opened 2015-10-29
Support stylish-haskell

bobjflong:
fixes https://github.com/Glavin001/atom-beautify/issues/628

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,3 +53,7 @@ before_install:
   - brew gist-logs php-cs-fixer || true
   # CoffeeScript
   - npm install coffee-formatter
+  # Haskell
+  - brew install haskell-stack
+  - stack setup
+  - stack install stylish-haskell
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ Or Settings/Preferences ➔ Packages ➔ Search for `atom-beautify`
 - [x] [Vala](https://github.com/Glavin001/atom-beautify/issues/57)
   - Requires [Uncrustify](http://sourceforge.net/projects/uncrustify/)
 - [x] [TypeScript](https://github.com/Glavin001/atom-beautify/issues/49)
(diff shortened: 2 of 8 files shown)


Glavin001:
This is great!

Could you also add a simple test example file and run the command `Window: Run package spec`. See step 4 from `How to add a Beautifier for a Language`: https://github.com/Glavin001/atom-beautify/blob/master/docs/add-languages-and-beautifiers.md#how-to-add-a-beautifier-for-a-language

You may have to edit [`.travis.yml`](https://github.com/Glavin001/atom-beautify/blob/master/.travis.yml#L30) and [`.appyveyor.yml`](https://github.com/Glavin001/atom-beautify/blob/master/appveyor.yml#L43) so the tests pass when run on CI.

bobjflong:
@Glavin001 I've added a test case. However I think that the only real way to install `stylish-haskell` is via Haskell source, so it might be more trouble than it's worth to get this working with CI :/

Glavin001:
> I've added a test case. 

Thanks!

> However I think that the only real way to install stylish-haskell is via Haskell source, so it might be more trouble than it's worth to get this working with CI :/

Bummer. If you can get it working on Travis CI that would be great -- I find Windows with AppVeyor to be too difficult to get working sometimes anyway, so Travis CI (Linux) is my Go-To CI for verifying Pull Requests.

If you are unable to get Travis CI working, then change the `original/test.hs` to `original/_test.hs` which will disable that test (tests with `_` prefix are ignored).

After that, I can merge this as is.

Thank you for contributing!

bobjflong:
I got travis to pass! Installing the hs dependencies adds to the travis build time but that's probably OK?

Glavin001:
Absolutely OK. This is great! Thank you for contributing.

Glavin001:
Published to v0.28.16

bobjflong:
:beers:

PR #636, opened 2015-11-02
Travis: cache stack builds

bobjflong:
In https://github.com/Glavin001/atom-beautify/pull/629 I added a stack build to Travis. This PR will cache the `~/.stack` dir and greatly speed up subsequent runs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ cache:
     - directories:
         - node_modules
         - $HOME/.atom
+        - $HOME/.stack
 
 before_install:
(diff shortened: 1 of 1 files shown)


Glavin001:
Thanks!

Glavin001:
Published to v0.28.17

bobjflong:
Ugh @Glavin001 I checked out why this change wasn't making much of a difference. Seems like Travis only support caching for private OSX builds (but public container based): http://docs.travis-ci.com/user/ci-environment/

PR #726, opened 2015-12-30
Added pretty diff ternaryline option - Issue 524

joshholl:
Added ternaryline option from pretty diff and corresponding documentation.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -1322,6 +1322,30 @@ Break chained method calls across subsequent lines (Supported by JS Beautify, Pr
 }
 ```
 
+####  [JavaScript - Preserve Ternary On Single Line](#javascript---preserve-ternary-lines) 
+
+**Namespace**: `js`
+
+**Key**: `preserve_ternary_lines`
+
+**Type**: `boolean`
+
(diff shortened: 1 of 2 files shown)


prettydiff:
Sorry for the delay.  I am reviewing your pull request right now.  I see that both JS Beautify and Pretty Diff are supported beautifiers for this option, but only the prettydiff.coffee file was modified to support this feature.

Does JS Beautify support this feature and if so does any additional code need to be changed to make this feature available to JS Beautify?  If JS Beautify does not support this enhancement it might be better to remove that beautifier from supporting this feature until they are ready.

jackocnr:
This is great! Do you know when this will be released? Thank you.

Glavin001:
Published to v0.28.23

jackocnr:
Thanks that's now working (using "preserve_ternary_lines" in my `.jsbeautifyrc` file). I'm now having some other problems, but I'll open a separate issue for that.

joshholl:
Thanks for the work on this, sorry i hadn't had a chance to respond to earlier messages.

PR #733, opened 2016-01-06
Adding dfmt support for D

LaurentTreguier:
This PR adds basic dfmt support.

### DIFF
--- a/src/beautifiers/dfmt.coffee
+++ b/src/beautifiers/dfmt.coffee
@@ -12,6 +12,6 @@ module.exports = class Dfmt extends Beautifier
   }
 
   beautify: (text, language, options) ->
-    @run("dfmt --indent_size 8", [
+    @run("dfmt", [
       @tempFile("input", text)
       ])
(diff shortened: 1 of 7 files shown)


Glavin001:
Awesome! Please add tests as well. See step 4 of `How to add a beautifier for a language`: https://github.com/Glavin001/atom-beautify/blob/master/docs/add-languages-and-beautifiers.md#how-to-add-a-beautifier-for-a-language

LaurentTreguier:
I added a simple test and put dfmt as default beautifier, since it seems better than uncrustify for beautifying D (with the code I tested them with at least), and also because the test would fail otherwise.

src/beautifiers/dfmt.coffee
```diff
@@ -0,0 +1,17 @@
...
+Requires [dfmt](https://github.com/Hackerpilot/dfmt)
+###
+"use strict"
+Beautifier = require('./beautifier')
+
+module.exports = class Dfmt extends Beautifier
+  name: "dfmt"
+
+  options: {
+    D: true
```
Glavin001: Should be `false`

Glavin001:
Published to v0.28.23

PR #743, opened 2016-01-13
Adds 'unchanged' option to the SQL beautifier

cosmix:
The sqlformat tool can preserve identifier and keyword case. This is particularly useful for identifiers in cases where the RDBMS is case sensitive. Arguably it might be useful for SQL keywords too. This patch enables this by adding another option ('unchanged') that does exactly that and defaults to unchanged in the case of identifiers.


Glavin001:
Published to v0.28.23

Issue #755, opened 2016-01-18
"if" clauses inside inline JavaScript on XHTML document formatting issue


keul:
The using the beautify feature on inline JavaScript I get some strange effects.
This is the HTML section with JS inside (manually beautified):

    <script type="text/javascript">
        $(document).ready(function ()  {
            if (mythema == '...') {
                foo();
                bar();
            }
        });
    </script>

This is what I get after:

    <script type="text/javascript">
        $(document).ready(function ()  {
            if (mythema == '...') {foo();
                bar();}
        });
    </script>

It doesn't make any sense to me.

prettydiff:
XHTML files should be handled more as XML than HTML type documents.  While XHTML documents contain the same vocabulary and rules as HTML they carry the syntax and well-formedness requirements of XML.  I have opened a pull request to resolve this issue.

This is not really an atom-beautify file defect, perhaps more of a missing feature (associating .xhtml files as XML documents).  I would also request you open an issue with the JS Beautify project as that is the default beautifier for HTML: https://github.com/beautify-web/js-beautify/issues

keul:
@prettydiff thanks for explanation.

Unluckily the file I'm editing is not even a XHTML because it's a template of the Zope application server (an XML syntax compatible language)

Maybe a way to disable prettifying in a section of the file can help? I can live with a not-well-indented section in my template, but not with a compressed mess on JavaScript.

prettydiff:
@keul Just select "XML" from the grammar list for this file (Edit, Select Grammar) and then when you beautify the problem will be resolved.

bitwiseman:
Also, I don't know how you're getting the output you say you're getting.   I don't know of any condition under which the output you're showing would happen.

prettydiff:
@keul Would you mind posting a gist of a complete sample template file so that we can attempt to reproduce this error?

keul:
@bitwiseman @prettydiff I'll try to gives you all details I can in next few days (I'm out of office until end of the week).
Thanks

bitwiseman:
@keul  - any update?

keul:
@bitwiseman I'm sorry, busy days and I totally forgot the pending issue.

Unluckily the issue... disappeared.
I copied the template file apart to perform additional tests and today it simply works, it's nicely beautified.
I'm not sure if recently the atom-beautify plugin has been updated so something fixed the issue or not.

Sorry to all for the noise.

@prettydiff: the text format was already XML also last time I checked.

[...]

PR #756, opened 2016-01-18
adding XHTML file extension as an XML grammar

prettydiff:
This request is a proposed resolution for https://github.com/Glavin001/atom-beautify/issues/755


Glavin001:
Published to v0.28.22

Issue #762, opened 2016-01-20
Support for SLD format


gallayc:
Standard format defined by OGC (Styled Layer Description) : formatter should be the same as XML formatter.

prettydiff:
This is complete and just awaiting publication.

Glavin001:
Published to v0.28.22

PR #763, opened 2016-01-20
adding support for grammar SLD

prettydiff:
Per https://github.com/Glavin001/atom-beautify/issues/762


Glavin001:
Published to v0.28.22

Issue #750, opened 2016-01-15
Feature: Add support for Riot.js tag


rodrigorodriguez:
Add support for Riot.js tag files: .tag, grammar HTML (Riot Tag).

prettydiff:
Duplicate https://github.com/Glavin001/atom-beautify/issues/731

This grammar appears to be supported by an included beautifier, so it just needs a pull request to associate `.tag` files with Pretty Diff beautifier.

prettydiff:
This is just pending the next release.

phillip-haydon:
Splooge! Hope new release is soooonnnnnn :D

rodrigorodriguez:
@prettydiff  That's what I meant as I wrote 'HTML (Riot Tag)', that's the name of the grammar.

Glavin001:
#766 has been merged.

Glavin001:
Published to v0.28.28

PR #766, opened 2016-01-22
Updating riotjs.coffee per comment: https://github.com/Glavin001/atom…

prettydiff:
Updating riotjs.coffee per comment: https://github.com/Glavin001/atom-beautify/issues/750#issuecomment-174029448


Glavin001:
Published to v0.28.23

Issue #776, opened 2016-02-01
Option to preserve inline object


Zephyrrus:
Not sure why it's happening now and never happened before to me, but the JS beautifier expands every object to a key:value per line, instead of keeping them in a single line. This is very annoying for short objects which could be kept on a single line but instead are expanded to several lines.

prettydiff:
The default beautifier for JavaScript is JS Beautify unless you change the beautifier in the Atom Beautify settings to Pretty Diff.  JS Beautify had a big update last week featuring more advanced object support and destructuring.

Would you be able to provide a sample of code so that we can produce the issue and indicate if you are using the default beautifier or Pretty Diff?

Zephyrrus:
I am using the default JS Beautify, and a simple example is 
```javascript
      "databaseStructure": [
        {name: "id", type: "autonumber", primaryKey: true},
        {name: "uid", type: "number", required: true},
        {name: "reason", type: "string"},
        {name: "addedDate", type: "datetime", required: true},
        {name: "addedBy", type: "number", required: true}
      ]
``` 
which get's expanded to 
```javascript
"databaseStructure": [
          {
              name: "id",
              type: "autonumber",
              primaryKey: true
          },
          {
              name: "uid",
              type: "number",
              required: true
          },
          {
              name: "reason",
              type: "string"
          },
          {
              name: "addedDate",
              type: "datetime",
              required: true
          },
          {
              name: "addedBy",
              type: "number",
              required: true
          }
      ]
``` 
after using JS beautify, and it's very annoying for me because I prefer to keep them on a single line.

prettydiff:
The output appears to be completely correct, though you are asking for a different stylistic preference.  I will include @bitwiseman so that he can weigh in if this is an enhancement that he would like to pursue for JS Beautify.

I just checked this in Pretty Diff and I am not supporting this preference either.  We both support object flattening, but only in certain contexts.  Universal object flattening is not something that has been requested to me and is generally not desired if the objects large enough to produce horizontal scrolling.

bitwiseman:
JSBeautify 1.6.x supports keeping inline objects by setting brace-style to `colapse-preserve-inline`.  

This input remains unchanged when that setting is enabled:
```js
{
    "databaseStructure": [
        { name: "id", type: "autonumber", primaryKey: true },
        { name: "uid", type: "number", required: true },
        { name: "reason", type: "string" },
        { name: "addedDate", type: "datetime", required: true },
[...]

PR #779, opened 2016-02-03
[#776] Bump jsbeautify to 1.6.2.  Add support for collapse-preserve-inline

kaaloo:
I have tested the package locally and the option seems to work fine now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - Add [elm-format](https://github.com/avh4/elm-format) beautifier for the Elm language.
 - Add [clang-format](http://clang.llvm.org/docs/ClangFormat.html) beautifier for C/C++/Obj-C languages.
 - Add [yapf](http://github.com/google/yapf) beautifier for Python.
+- Closes [#776] (https://github.com/Glavin001/atom-beautify/issues/776) Add support for `collapse-preserve-inline` brace_style for javascript.
 
 # v0.29.0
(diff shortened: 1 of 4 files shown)


lnewman-book:
Failed tests?

prettydiff:
Yeah, the unit tests appear to need some work.

lnewman-book:
Could I ask that you hold off on this for today?  There's been a request for moving `preserve-inline` to a separate setting to allow it to be combined with `expand` and `end-expand`.  If that happens, it would cause churn in this area.  Let me see how difficult that would be.

bitwiseman:
Please hold of on this for today.  Looking at options making this option less disruptive.

bitwiseman:
Hmm, never mind.  I don't have the bandwidth for further changes at this time.

kaaloo:
Yup, probably because of the jsbeautify version bump.

ttsirkia:
+1 for this. It would be great to finally get this particular feature.

NerdPad:
+1 if we can get this feature.

kaaloo:
Thank you that's awesome!

Glavin001:
Published to v0.28.23

ttsirkia:
After a short experiment, it seems that objects are sitll expanded to multiple lines as before although this new option is selected. Can someone else share experiences?

Glavin001:
@ttsirkia please create a new issue and follow the instructions provided in the template. Thank you.

ttsirkia:
Sure, #831. I was just asking if someone could give a quick comment if there is only something wrong in my environment.

PR #790, opened 2016-02-09
Rust - Add ability to read rustfmt.toml

sidred:
This is a breaking change for existing rust format users. rustfmt version 0.3.0 or newer is required. 

The older version of rustfmt will error with the message
```
Unrecognized option: 'config-path'.
```

This commit 
- adds the --config-path parameter with the current folder and will load any rust format in the file's path.
- sets the --write-output format to override. This will avoid creating an unnecessary backup of the tmp source file


sidred:
Not sure why travis failed. Does not seem to be related to the rustfmt changes.

Glavin001:
Published to v0.28.23

PR #791, opened 2016-02-09
Wraps .jsbeautifyrc file read into a try/catch since it could be no l…

diegodorado:
…onger available.


Glavin001:
Published to v0.28.23

PR #832, opened 2016-02-26
Indent javascript using D3 conventions

mindrones:
Hi, this patch is not meant for inclusion as it is but rather to show what I'm trying to achieve and to ask for direction.

In `d3`, by convention one should indent differently based on if the kind of selection returned by a call to the selection API.

From https://github.com/d3/d3-selection#d3-selection:

> By convention, selection methods that return the current selection use *four* spaces of indent, while methods that return a new selection use only *two*. This helps reveal changes of context by making them stick out of the chain:
>
> ```js
d3.select("body")
  .append("svg")    // 2 because `append` returns a new selection (svg)
    .attr("width", 960)    // 4 because `attr` modifies the current selection (svg)
    .attr("height", 500)    // idem 
  .append("g")    // 2 because `append` returns a new selection (g)
    .attr("transform", "translate(20,20)")    // ...and so on
  .append("rect")
    .attr("width", 920)
    .attr("height", 460);
```

In this first experiment I'm simply doing some replacement after beautifying with `js-beautify` but I'm sure this is not the right way to do this:
- is it possible to do what I want?
- if so, could you kindly shed some light on how to do this?

Specifically I'd like to target only "D3 calls" so I'm aware I'll need to also parse and apply these changes only in D3 "context".

Thanks!

### CODE REVIEW

**Comment by prettydiff** [comment]
This would be an enhancement to JSBeautify, Pretty Diff, or both.

There are some technical challenges in this approach:

* Indentation is open to customization.  It can be composed of a variable number of spaces, tabs, or possibly something else.  Determining half and indentation becomes more challenging when indentation is defined as a single character.
* The beautifiers would have to account for all the D3 methods and determine which ones require special treatment.

**Comment by mindrones** [comment]
@prettydiff thanks.

> Determining half and indentation becomes more challenging when indentation is defined as a single character.

Agree: maybe method calls returning a new selection (RANS) can get indented of one indentation "block", while method calls not returning a new selection (NRANS) get indented by one indentation "block" + 2 blanks or by 2 indentation blocks?

> The beautifiers would have to account for all the D3 methods and determine which ones require special treatment.

Well only [d3 selection](https://github.com/d3/d3-selection) methods, which aren't many.
Maybe these can be listed as a new mini grammar?

Also, if another library uses same method names, we should indent "more" only method calls NRANS chained to objects which are `instanceof d3.selection`:

```js
var body = d3.select('body')
body
  .append('div')
    .attr('foo', 'bar')

var data = mylib(mydata)
data
  .select('foo')
  .append('bar')
```

PR #856, opened 2016-03-14
Fix yapf style issue

### LINKED ISSUE #786: YAPF configuration files are ignored
YAPF looks for certain [configuration files](https://github.com/google/yapf#formatting-style) containing user preferences regarding coding formatting. I have these configuration files in two places: `~/.config/yapf/style` and a `.style.yapf` file in my project directory. However, when I run Atom Beautify in a Python file, using YAPF as my default formatting engine, it seems to ignore these files. I tried to setup YAPF from Atom Beautify settings page from inside Atom, but it only has settings for `autopep8`, apparently.

Here's my [debug info](https://gist.github.com/douglasrizzo/075d005027c9ade753f0). I configured YAPF to only break lines with more than 150 characters, but it is stilll breaking lines with ore than 79 characters, as it is its default.

kdheepak89:

YAPF configuration files were previously being ignored. This was because the `run` call to yapf was explicitly defining a `--style` flag.

The --style flag accepts different types of input. From their documentation - "accepts one of the predefined styles (e.g., pep8 or google), a path to a configuration file that specifies the desired style, or a dictionary of key/value pairs."

By not explicitly specifying the --style flag, yapf will look for the following in order

1. In the [style] section of a .style.yapf file in either the current directory or one of its parent directories.
2. In the [yapf] secionf of a setup.cfg file in either the current directory or one of its parent directories.
3. In the ~/.config/yapf/style file in your home directory.

This pull request removes the `--style=pep8` flag in the `run` call to yapf. This is not required, since yapf uses `pep8` by default

This closes #786 

### Any other comments?

A further improvement will be to have a field where the style can be entered as a dictionary.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Add [clang-format](http://clang.llvm.org/docs/ClangFormat.html) beautifier for C/C++/Obj-C languages.
 - Add [yapf](http://github.com/google/yapf) beautifier for Python.
 - Closes [#776] (https://github.com/Glavin001/atom-beautify/issues/776) Add support for `collapse-preserve-inline` brace_style for javascript.
+- Closes [#786](https://github.com/Glavin001/atom-beautify/issues/786) YAPF configuration files are ignored.
 
 # v0.29.0
 - Closes [#447](https://github.com/Glavin001/atom-beautify/issues/447). Improved Handlebars language support
--- a/package.json
+++ b/package.json
@@ -70,6 +70,10 @@
     {
       "name": "Bati Sengul",
       "url": "https://github.com/Focus"
+    },
+    {
+      "name": "Dheepak Krishnamurthy",
+      "url": "https://github.com/kdheepak89"
     }
   ],
   "engines": {
--- a/src/beautifiers/yapf.coffee
+++ b/src/beautifiers/yapf.coffee
@@ -16,7 +16,6 @@ module.exports = class Yapf extends Beautifier
   beautify: (text, language, options) ->
     @run("yapf", [
       "-i"
-      ["--style=pep8"]
       tempFile = @tempFile("input", text)
       ], help: {
         link: "https://github.com/google/yapf"


ProgramFan:
The patch is clear enough to understand. However, there are still design decisions to made. When I introduced the yapf beautifier, I assume it an alternative for pep8, so the '--style=pep8' was hardcoded. It looked a safe choice at that time.

Now we want to support the yapf mechanism for style config. There is an additional issue to solve before this can get merged: the file passed to 'yapf' command is an temporary file, usually created at /tmp/xxx, not in the working directory. This can force yapf to read ~/.config/yapf/style in all cases, ignoring any settings in option 1 and 2 above.

The same problem happens to clang-format beautifier and gets solved by maintaining a seperate temporary file under the same directory as the original file. You can refer to the clang-format beautifier for more details.

kdheepak89:
That makes sense. Would a suitable alternative be to copy the file for option 1 and 2 if it exists to the tmp folder? Also, in the current implementation, we shouldn't have the `--style=pep8` hardcoded because pep8 is the default for yapf. This limits use of all options. With the line removed, users can at the very least use a global style file.

ProgramFan:
I agree. It makes sense to copy the style file to the temp dir. The problem to watch out is to make sure the two temp files get created under the same directory and to search the config file as least as possible (I guess search for a filename in a large directory can be slow thus cause possible lag).

Glavin001:
Please let me know when this Pull Request is complete and ready for review :smiley:. Thanks in advance for contributing!

kdheepak89:
Hi @Glavin001, I believe that this pull request is ready for review at its current state and if merged will be beneficial for users. However, I do agree with @ProgramFan and think it can be improved, but I'm currently not sure how to implement those changes.

Glavin001:
Oh looks like some new changes are causing merge conflicts.

If you could update this branch to merge the latest master branch then I will merge this in and so it will be included in the next release. Thanks!

kdheepak89:
Hi @Glavin001, thanks for looking into this pull request! I didn't realize there were merge conflicts (I was on my mobile). I've updated the branch and it should be now ready to merge with `master`

Glavin001:
No problem, it's understandable.

I am going to merge this now. If there are any future improvements please submit as another Pull Request. Thanks!

Glavin001:
Published to v0.28.28

ostrokach:
How difficult would it be to add the ability to use configurations in the package `setup.cfg` file? Judging by the comments above, `clang-format` does this already?

ProgramFan:
Yes. The clang-format beautifier achieves this by creating explicitly a temporary file in the same directory as the current file and feed it into clang-format, so clang-format find the correctly .clang-format file as if it is doing this for the original file. The major job here is to implement the temporary file creation and deletion, both of which have to be done with extra code.

PR #864, opened 2016-03-21
[WIP] Better settings and performance

### LINKED ISSUE #713: Beautify settings UX
Hello,

I like this plugin and find it really useful!

But as for me, the current state of settings is really messy. First, at the top you have a list of settings for every language. Not really categorized, "Language-Name - " prefixes are used instead of nice, clean sections:

screen shot 2015-12-16 at 4 21 53 pm

But then goes the worst part of UX: when you scroll down, you find another list of settings for every language again:

screen shot 2015-12-16 at 4 22 11 pm

Again, prefixes are used instead of sectioning.

Here are some ideas how to improve UX:
* make clean sections for every language, divided by headers with its name
* do either of:
  * merge all settings concerning one language into one section
  * make two main sections from what we have now

Please, tell me what you think about it. I really think that this nice plugin deserves a nicer look of settings.

I'm new to Atom plugins development, but I think I could create a pull request some time if nobody else would do that.

Glavin001:

Switch to using `object` type for options in Atom-Beautify's package settings.
Thus allowing for collapsable setting groups. See https://github.com/atom/settings-view/pull/736 for more details.

Here is what it looks like now, all organized and groups collapsed:
image

Then you can expand all of the groups:
image

Closes #713.

### Any other comments?

I am very excited for this. However, it will be a breaking change.
I hope to have built-in migrate so this will be seamless for users.

### Checklist

---

Todo:
- [x] Options are now in `object` groups for each language
  - [x] Language config options within language groups
  - [x] Include options from `fallback` languages
  - [x] Delete options that do not have a supported beautifier
- [x] Options are being extracted (tests all passing)
  - [x] Add tests for using the package settings instead of `.jsbeautifyrc`
- [x] Documentation generation is updated to handle this new structure
- [x] Options are migrated automatically from `{ LANG_OPTION : 'VALUE' }` to `{ LANG: {OPTION:'VALUE'}}` format
- [x] Pre-compile the language/beautifiers config settings


Glavin001:
I also want to fix #876 with this Pull Request.

Glavin001:
Published to v0.29.0

PR #916, opened 2016-04-13
allow beatify *.pug files

ewgenius:
It seems, what default extension of jade now is `pug`, but atom-beautifier (with `language-pug` package) can't handle `*.pug` files


Glavin001:
Thank you for contributing!

Glavin001:
Published to v0.29.3

Issue #707, opened 2015-12-13
Erroneously changes CRLF to LF on Windows


garretwilson:
atom-beautify at first seemed nice, but then I realized it changed all my CRLF line endings on Windows to LF. How do I turn this off?

This is a bug. I shouldn't have to find a setting to keep my current CRLF line endings. atom-beautify should stay with my current line endings; or at the very, very least detected my system and use appropriate line endings. The current behavior is incorrect as a default.

garretwilson:
This is a huge blocker issue. I see that someone else opened the same issue (#593) for JavaScript, and it was never resolved.

I can't have something mucking around with my line endings. This messes up my workflow, especially with Git. (I am well-versed in `.gitattributes`, etc.) This is a bug. It needs fixed. Until then, I'm dumping atom-beautify. Not to be mean, but because it's messing up my files.

garretwilson:
I don't know which side this bug is on, so I have [filed it on js-beautify](https://github.com/beautify-web/js-beautify/issues/829).

prettydiff:
The default beautifier for JavaScript is JS Beautify.  If they don't have an issue open on this then open an issue there.  We do not control which options they choose to accept or when those options are completed.

Pretty Diff added this feature with https://github.com/prettydiff/prettydiff/issues/186 and the addition of this option to Atom Beautify is being tracked at: https://github.com/Glavin001/atom-beautify/issues/524

garretwilson:
What is odd is that the [Komodo Beautify js add-on](https://github.com/babobski/Beautify-js) doesn't change the line endings. I wonder how Komodo can use Beautify without ruining my line endings, yet atom-beautify cannot?

prettydiff:
The difference is that the line termination sequence is designated by the IDE and not the beautifier.  The beautifier must just be willing to accept an optional definition for the line termination sequence and apply it appropriately.  Many IDEs are like this in that they allow the user to choose an accept line termination sequences from a list for any given file.  I have not used Komodo Edit so I don't know if it has the capability.  Notepad++ is a Windows only editor that allows the user to change line termination sequence in nearly the same way as Atom.

garretwilson:
I'm not understanding you. Are you saying that Atom has no clue about what line-ending is being used for a particular file, and therefore cannot pass this option to JS Beautify?

prettydiff:
[...]

PR #941, opened 2016-04-23
Fix issue #707 by adding Atom-based EOL detection

garretwilson:
Fixes issue #707 by adding Atom-based EOL detection.

Closes issue #707 (which was already closed through miscommunication between projects).

### Any other comments?
Determines the default line ending based upon the Atom configuration
`line-ending-selector.defaultLineEnding`. If the Atom configuration
indicates "OS Default", the `process.platform` is queried, returning
CRLF for Windows systems and LF for all other systems.

### DIFF
--- a/src/beautifiers/js-beautify.coffee
+++ b/src/beautifiers/js-beautify.coffee
@@ -24,6 +24,9 @@ module.exports = class JSBeautify extends Beautifier
   beautify: (text, language, options) ->
     @verbose("JS Beautify language #{language}")
     @info("JS Beautify Options: #{JSON.stringify(options, null, 4)}")
+    #TODO reconsider handling of EOL once js-beautify adds EOL detection
+    #see https://github.com/beautify-web/js-beautify/issues/899
+    options.eol = getDefaultLineEnding() ? options.eol #fixes issue #707
     return new @Promise((resolve, reject) =>
       try
         switch language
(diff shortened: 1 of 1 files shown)


jerone:
This appears to be not backwards-compatible with build-in [line-ending-selector](https://github.com/atom/line-ending-selector) status-bar package.

Can I suggest an fourth option: `Auto` (or something similar). Which detects the current line-ending per file. [Here are some hints](https://github.com/atom/line-ending-selector/blob/master/lib/main.js#L122-L128) to get this in.

garretwilson:
> This appears to be not backwards-compatible with build-in line-ending-selector status-bar package.

Please explain specifically what you mean by "not backwards-compatible".

1. Describe your current configuration/scenario.
2. Explain the behavior that happens with this pull request.
3. Indicate the behavior you think should happen instead.

Thanks.

Glavin001:
The failing test is unrelated. Looks good to merge. Thank you for contributing! 😃

Glavin001:
Any improvements to this feature can be submitted in a new Pull Request. Thanks!

jerone:
> Please explain specifically what you mean by "not backwards-compatible".

The official [line-ending-selector](https://github.com/atom/line-ending-selector) allows you to change the line-endings per file. This pull request just checks the current OS default line-ending, which in cross-platform projects may not be true if you have your git checkout line-endings defined as-is.

My suggestion was to actual [check the current file](https://github.com/atom/line-ending-selector/blob/master/lib/main.js#L125) for the used line-endings before "forcing" a line-ending.

garretwilson:
Which "line ending setting" is used is a larger issue open to discussion and no doubt differences of opinion. It would be difficult to say that one is right or the other wrong, but rest assured that I did consider whether the current file line ending should be taken into account. I believe the approach I used is correct, and I outline my reasoning below.

As you note the line-ending-selector package indicates the _actual_ line-ending of the _current_ file. The whole purpose of a "beautifier" or "formatter", however, is based on the assumption that the "current" format may not be correct; the beautifier therefore will "normalize" the current file into some "canonical" format.

The question then becomes: what is the "canonical" format that should be used? What better way is it to determine the canonical format than to see which setting the user has used for the Atom configuration?

You mention as a drawback the situation in which a user has configured Git (e.g. using the `core.autocrlf` property or via a `.gitattributes` file) to use some fixed line ending and to ignore the system default EOL delimiter (to standardize on LF across platforms, for example). But surely in such a situation, the user would have configured the line-ending-selector package  so that instead of "OS Default" it is "LF" or "CRLF".

Put another way, js-beautify should not be checking the setting of the _current_ file or the _current_ setting of the line-ending-selector, but instead its configured _default_ value---and in fact my pull request does exactly that.

Glavin001:
Published to v0.29.7

PR #956, opened 2016-04-29
Marko v3

patrick-steele-idem:

The Marko templating language had a new major release that introduced a new syntax. Marko now parses all HTML attributes as JavaScript expressions and it supports both an HTML syntax and a concise syntax. See: [What's New in Marko v3](http://markojs.com/docs/marko/what-is-new-marko-v3/)

No.

### Any other comments?

The Marko beautifier was updated to use the following package: [marko-prettyprint](https://github.com/marko-js/marko-prettyprint). The `marko-prettyprint` tool needs to know the file system path associated with the text being beautified since custom tags need to be resolved relative to the location of the template on disk and tag definitions can impact pretty printing (e.g., some tags are declared to preserve whitespace). The `atom-beautifier` plugin was not providing the file path so I introduced a new `context` argument that now includes the `filePath` property (if available) as a fourth argument. The new signature is: `beautify(text, language, options, context)`

### Checklist

- [X] Added examples for testing to [examples/ directory](examples/)

### DIFF
--- a/src/beautifiers/js-beautify.coffee
+++ b/src/beautifiers/js-beautify.coffee
@@ -9,7 +9,6 @@ module.exports = class JSBeautify extends Beautifier
     XML: true
     Handlebars: true
     Mustache: true
-    Marko: true
     JavaScript: true
     JSON: true
     CSS:
@@ -41,7 +40,7 @@ module.exports = class JSBeautify extends Beautifier
             beautifyHTML = require("js-beautify").html
             text = beautifyHTML(text, options)
             resolve text
-          when "HTML (Liquid)", "HTML", "XML", "Marko", "Web Form/Control (C#)", "Web Handler (C#)"
+          when "HTML (Liquid)", "HTML", "XML", "Web Form/Control (C#)", "Web Handler (C#)"
             beautifyHTML = require("js-beautify").html
             text = beautifyHTML(text, options)
             @debug("Beautified HTML: #{text}")
--- a/src/languages/marko.coffee
+++ b/src/languages/marko.coffee
@@ -1,8 +1,15 @@
+# Get Atom defaults
+scope = ['text.marko']
+tabLength = atom?.config.get('editor.tabLength', scope: scope) ? 4
+softTabs = atom?.config.get('editor.softTabs', scope: scope) ? true
+defaultIndentSize = (if softTabs then tabLength else 4)
+defaultIndentChar = (if softTabs then " " else "\t")
+defaultIndentWithTabs = not softTabs
+
 module.exports = {
 
   name: "Marko"
   namespace: "marko"
-  fallback: ['html']
 
   ###
   Supported Grammars
@@ -18,6 +25,22 @@ module.exports = {
     "marko"
   ]
 
-  options: []
+  options:
+    indent_size:
+      type: 'integer'
+      default: defaultIndentSize
+      minimum: 0
+      description: "Indentation size/length"
+    indent_char:
+      type: 'string'
+      default: defaultIndentChar
+      description: "Indentation character"
+    syntax:
+      type: 'string'
+      default: "html"
+      enum: ["html", "concise"]
+      description: "[html|concise]"
+
+  defaultBeautifier: "Marko Beautifier"
 
 }
--- a/src/languages/marko.coffee
+++ b/src/languages/marko.coffee
@@ -10,6 +10,7 @@ module.exports = {
 
   name: "Marko"
   namespace: "marko"
+  fallback: ['html']
 
   ###
   Supported Grammars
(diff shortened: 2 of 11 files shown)


patrick-steele-idem:
Hey @Glavin001, sorry to bug you, but just wanted to follow up on this Pull Request. Do you have any questions, thoughts or concerns? If you are too busy to review right now then no worries. Thanks for maintaining this project!

src/languages/marko.coffee
```diff
@@ -1,8 +1,15 @@
...
+softTabs = atom?.config.get('editor.softTabs', scope: scope) ? true
+defaultIndentSize = (if softTabs then tabLength else 4)
+defaultIndentChar = (if softTabs then " " else "\t")
+defaultIndentWithTabs = not softTabs
+
 module.exports = {
 
   name: "Marko"
   namespace: "marko"
-  fallback: ['html']
```
Glavin001: Are you sure you do not want to fallback to `html` settings? The settings in the `makro` namespace will take precedence but in the case that `makro` is not set then `html` will be used.

Glavin001:
Hey @patrick-steele-idem! Thanks for reminding me.
I've been a little busy with work, however I have looked over this PR.

My only concern is adding the `filePath`. Although, from your comment above, I do agree it is necessary.
In the case that a user beautifies an unsaved file -- such that there is no `filePath` -- then will Makro throw an error or just beautify with unexpected formatting because it does not know the tag definitions?

patrick-steele-idem:
> then will Makro throw an error or just beautify with unexpected formatting because it does not know the tag definitions?

The Marko beautifier still works if file path is not present since it will just use a default directory. Without a path, custom tags (if any) won't be resolved correctly and that might have a small impact on the beautified content, but it will likely never be noticed as an issue.

Thanks for taking the time to review the code. As a side note, I had a hard time running the tests and they took a really long time on Travis CI and AppVeyor. I think the Marko tests are passing and that there are other tests that are failing, but I am not entirely sure.

src/languages/marko.coffee
patrick-steele-idem: Marko templates no longer use the HTML syntax (Marko uses a variation of HTML that is not always valid HTML). Therefore, you would not be able to use the default HTML beautifier to beautify a Marko v3 template. Also, currently the Marko beautifier doesn't support the majority of the settings that the HTML beautifier supports. That's my thought process for removing `html` as a fallback, but please let me know if I am mistaken. Thanks again!

Glavin001: > Marko templates no longer use the HTML syntax (Marko uses a variation of HTML that is not always valid HTML). Therefore, you would not be able to use the default HTML beautifier to beautify a Marko v3 template. 

That's not what `fallback` does 😃. `fallback` is only related to how to get configuration settings.

> Also, currently the Marko beautifier doesn't support the majority of the settings that the HTML beautifier supports. That's my thought process for removing html as a fallback,

Leaving `html` as a fallback means that a user could set `html.indent_size` and if `marko.indent_size` is not set then it will look at `html.indent_size` next, etc. So the options `indent_size` and `indent_char` would be applicable, however `syntax` would not.

Fallback option was essentially a way to allow users to configure for one main language (`html`) and similar languages (`marko`?) could then take advantage of those settings without requiring explicitly changing the settings for that language itself (`marko.indent_size` directory).

src/beautifiers/js-beautify.coffee
```diff
@@ -9,7 +9,6 @@ module.exports = class JSBeautify extends Beautifier
     XML: true
     Handlebars: true
     Mustache: true
-    Marko: true
```
Glavin001: Does `js-beautify` handle Marko language well enough? If so, then leaving Marko support in JS-Beautify is a good idea, so users can choose and switch between them. We then set `defaultBeautifier` for the language to `Marko Beautifier` instead.

patrick-steele-idem: Hey @Glavin001, sorry for not responding earlier, but I took a vacation. 

> Does js-beautify handle Marko language well enough?

I would say that, no, `js-beautify` does not handle the Marko v3 syntax well enough. Marko allows HTML attributes to be arbitrary JavaScript expressions and that will often completely throw off `js-beautify` and likely break the code.

src/languages/marko.coffee
patrick-steele-idem: I see...my understanding was wrong. Thanks for clarifying! I have added `html` back as a fallback. Do you have any other concerns or see any other outstanding issues? Thanks, again, for all your help and maintaining this project.

Glavin001:
Thank you for contributing!

patrick-steele-idem:
Thanks @Glavin001 and thanks again for maintaining this project!

Glavin001:
Published to v0.29.8

PR #960, opened 2016-04-30
Enable use of JavaScript beautifier with JXA files

RobTrew:
Enable use of JavaScript beautifier with (OS X) JavaScript for
Automation (JXA) files.

JXA files use the extensions .scpt and .applescript

See package: language-javascript-jxa

### DIFF
--- a/src/languages/javascript.coffee
+++ b/src/languages/javascript.coffee
@@ -15,14 +15,17 @@ module.exports = {
   Supported Grammars
   ###
   grammars: [
-    "JavaScript"
+    "JavaScript",
+    "JavaScript for Automation (JXA)"
   ]
 
   ###
   Supported extensions
   ###
   extensions: [
-    "js"
+    "js",
+    "scpt",
+    "applescript"
   ]
 
   defaultBeautifier: "JS Beautify"
--- a/src/languages/javascript.coffee
+++ b/src/languages/javascript.coffee
@@ -24,8 +24,10 @@ module.exports = {
   ###
   extensions: [
     "js",
-    "scpt",
-    "applescript"
+	"scpt"
+	"jxa"
+	"jxainc"
+	"applescript"
   ]
 
   defaultBeautifier: "JS Beautify"


src/languages/javascript.coffee
```diff
@@ -15,14 +15,17 @@ module.exports = {
...
   ]
 
   ###
   Supported extensions
   ###
   extensions: [
-    "js"
+    "js",
+    "scpt",
+    "applescript"
```
danielbayley: I don't think this should work on _.applescript_ files.
[Here](https://github.com/danielbayley/atom-language-javascript-jxa/blob/master/grammars/jxa.cson#L3-L7) are the `fileTypes:` I have in my _language-javascript-jxa_ package… so this should probably match those.

Glavin001: :+1: I agree. @RobTrew please fix the `extensions` supported. Thanks.

RobTrew: Thanks – I have now added the full list.

	'scpt'
	'jxa'
	'jxainc'
	'applescript'

Glavin001:
2 issues:
- [ ] Support for JXA should really be in a separate Atom-Beautify language, called JXA, not in the JavaScript language
- [ ] Need examples of JXA with passing tests

RobTrew:
You think so ? "JavaScript for Automation" (JXA is not Apple's term - unclear where it comes from) is simply Safari JavaScript, and uses the Safari debugger.

(The only difference is an additional library in the global context)

danielbayley:
> Support for JXA should really be in a separate Atom-Beautify language, called JXA, not in the JavaScript language

> is simply Safari JavaScript, and uses the Safari debugger.

Exactly; it's literally just JavaScript, without any difference in syntax as far as I can tell. I wouldn't want to maintain another package anyway.

> JXA is not Apple's term - unclear where it comes from

It's a silly name, but it _is_ Apple's term. See [here](https://developer.apple.com/videos/play/wwdc2014/306) and [here](https://developer.apple.com/library/mac/releasenotes/InterapplicationCommunication/RN-JavaScriptForAutomation/Articles/Introduction.html).

src/languages/javascript.coffee
```diff
@@ -15,14 +15,19 @@ module.exports = {
...
+    "JavaScript for Automation (JXA)"
   ]
 
   ###
   Supported extensions
   ###
   extensions: [
-    "js"
+    "js",
+	"scpt"
```
Glavin001: Is the indentation wrong here?
Travis CI is failing: https://travis-ci.org/Glavin001/atom-beautify/jobs/130276709

Glavin001:
> it's literally just JavaScript

Sounds good. Keep support in the same Atom-Beautify language file, JavaScript.

My main issue is the inclusion of `applescript` and `scpt` extensions, since they are not always JavaScript and can be AppleScript. I suppose this difference is detected by https://github.com/danielbayley/atom-language-javascript-jxa/blob/master/grammars/jxa.cson#L9 and enables the appropriate grammar.

I still need at least 1 test script and make sure the relevant tests are passing. Thanks!

danielbayley:
> My main issue is the inclusion of applescript and scpt extensions, since they are not always JavaScript and can be AppleScript. I suppose this difference is detected by https://github.com/danielbayley/atom-language-javascript-jxa/blob/master/grammars/jxa.cson#L9 and enables the appropriate grammar.

Poorly thought out by Apple. That regex was the best I could do to try and differentiate them.

Issue #358, opened 2015-05-22
Support for Rust


fabianmeyer:
Please add support for the Rust language. Rust 1.0 was released on 05-15-2015 and the syntax is stable now.

nicokoch:
[https://github.com/nrc/rustfmt](https://github.com/nrc/rustfmt) might be helpful

Glavin001:
Excellent, @nicokoch! That looks great.

For those interested in making this happen, you can submit a Pull Request. I am still working on complete documentation however I have documented most of the steps required to add a new Language and a new Beautifier: https://github.com/Glavin001/atom-beautify/issues/300#issuecomment-107590998

Glavin001:
I have started development on the Rustfmt beautifier support however I have not tested it and it likely is not in a working state, yet. Feel free to fork Atom Beautify, and try it out for yourself and submit a Pull Request when it's completely working. Thanks.

Glavin001:
**Update**: I will be maintaining this documentation on adding languages and beautifiers: https://github.com/Glavin001/atom-beautify/blob/master/docs/add-languages-and-beautifiers.md

mattmaslin:
Rustfmt works correctly.  The only extra step I had to do was copy the <a href="https://github.com/nrc/rustfmt/blob/master/src/default.toml">default.toml</a> file from the Rustfmt project into my home directory.

LaylConway:
It seems there's a problem with the way rustfmt is used by Beautify. When trying to do a format, I get the following error.
```
C:\Users\Layl\AppData\Local\Temp\tmp1151030-5452-fkq6dy:5:5: 5:13 error: file not found for module `frontend`
C:\Users\Layl\AppData\Local\Temp\tmp1151030-5452-fkq6dy:5 mod frontend;
                                                              ^~~~~~~~
C:\Users\Layl\AppData\Local\Temp\tmp1151030-5452-fkq6dy:5:5: 5:13 help: name the file either frontend.rs or frontend/mod.rs inside the directory "C:\\Users\\Layl\\AppData\\Local\\Temp"
thread '<main>' panicked at 'Box<Any>', C:\Users\Layl\.cargo\git\checkouts\syntex-b2c5d63b3d829888\master\syntex_syntax\src\parse/mod.rs:76
```

It seems that rustfmt is expecting the entire cargo project to be in the location it wants to format it.

Drakulix:
[...]

PR #975, opened 2016-05-16
Severely improved support for rustfmt 0.5

lhecker:

`rustfmt` 0.5 contains an important but [not that obvious fix](https://github.com/rust-lang-nursery/rustfmt/pull/966). This PR allows us to directly feed the editor contents into `rustfmt` from `stdin` and read the beautified contents from `stdout`.

This not just severly improves performance on slower harddrives, but also fixes a very annoying bug with atom-beautify:
Up till now you could not beautify any files containing Rust's module syntax (`(pub) mod some_file_or_directory_name`), since the referenced files can obviously only be found in the actual working directoy and not in the temporary folder used by `atom-beautify`.

#358 or specifically https://github.com/Glavin001/atom-beautify/issues/358#issuecomment-160482164

### Any other comments?

- To make writing/reading work from stdin/stdout I had to add some facilities to `class Beautifier` to support that (it seems to me that all other beautifiers use temporary files?). I'd be cool if someone could take a look at it.
- Beautify on save will spuriously fail, but AFAIR this happened even before my change.

### Checklist

### DIFF
--- a/src/beautifiers/beautifier.coffee
+++ b/src/beautifiers/beautifier.coffee
@@ -113,7 +113,7 @@ module.exports = class Beautifier
   _envCacheDate: null
   _envCacheExpiry: 10000 # 10 seconds
   getShellEnvironment: ->
-    return new @Promise((resolve, reject) =>
+    return new Promise((resolve, reject) =>
       # Check Cache
       if @_envCache? and @_envCacheDate?
         # Check if Cache is old
@@ -248,71 +248,78 @@ module.exports = class Beautifier
   ###
   Run command-line interface command
   ###
-  run: (executable, args, {ignoreReturnCode, help} = {}) ->
+  run: (executable, args, {cwd, ignoreReturnCode, help, onStdin} = {}) ->
     # Flatten args first
     args = _.flatten(args)
+
     # Resolve executable and all args
     Promise.all([executable, Promise.all(args)])
-    .then(([exeName, args]) =>
-      @debug('exeName, args:', exeName, args)
-      return new Promise((resolve, reject) =>
+      .then(([exeName, args]) =>
+        @debug('exeName, args:', exeName, args)
+
         # Remove undefined/null values
         args = _.without(args, undefined)
         args = _.without(args, null)
+
         # Get PATH and other environment variables
-        Promise.all([@getShellEnvironment(), @which(exeName)])
-        .then(([env, exePath]) =>
-          @debug('exePath, env:', exePath, env)
-          exe = exePath ? exeName
-          # Spawn command
-          options = {
-            env: env
-          }
-          cmd = @spawn(exe, args, options)
-            .then(({returnCode, stdout, stderr}) =>
-              @verbose('spawn result', returnCode, stdout, stderr)
-              # If return code is not 0 then error occured
-              if not ignoreReturnCode and returnCode isnt 0
-                err = new Error(stderr)
-                windowsProgramNotFoundMsg = "is not recognized as an \
-                internal or external command" # operable program or batch file
-                @verbose(stderr, windowsProgramNotFoundMsg)
-                if @isWindows and returnCode is 1 and \
-                stderr.indexOf(windowsProgramNotFoundMsg) isnt -1
-                  err = @commandNotFoundError(exeName, help)
-                reject(err)
-              else
-                resolve(stdout)
-            )
-            .catch((err) =>
-              @debug('error', err)
-              # Check if error is ENOENT
-              # (command could not be found)
-              if err.code is 'ENOENT' or err.errno is 'ENOENT'
-                reject(@commandNotFoundError(exeName, help))
-              else
-                # continue as normal error
-                reject(err)
-            )
-        )
+        Promise.all([exeName, @getShellEnvironment(), @which(exeName)])
       )
-    )
+      .then(([exeName, env, exePath]) =>
+        @debug('exePath, env:', exePath, env)
+
+        exe = exePath ? exeName
+        options = {
+          cwd: cwd
+          env: env
+        }
+
+        @spawn(exe, args, options, onStdin)
+          .then(({returnCode, stdout, stderr}) =>
+            @verbose('spawn result', returnCode, stdout, stderr)
+
+            # If return code is not 0 then error occured
+            if not ignoreReturnCode and returnCode isnt 0
+              # operable program or batch file
+              windowsProgramNotFoundMsg = "is not recognized as an internal or external command"
+
+              @verbose(stderr, windowsProgramNotFoundMsg)
+
+              if @isWindows and returnCode is 1 and stderr.indexOf(windowsProgramNotFoundMsg) isnt -1
+                throw @commandNotFoundError(exeName, help)
+              else
+                throw new Error(stderr)
+            else
+              stdout
+          )
+          .catch((err) =>
+            @debug('error', err)
+
+            # Check if error is ENOENT (command could not be found)
+            if err.code is 'ENOENT' or err.errno is 'ENOENT'
+              throw @commandNotFoundError(exeName, help)
+            else
+              # continue as normal error
+              throw err
+          )
+      )
 
   ###
   Spawn
   ###
-  spawn: (exe, args, options) ->
+  spawn: (exe, args, options, onStdin) ->
     return new Promise((resolve, reject) =>
       @debug('spawn', exe, args)
+
       cmd = spawn(exe, args, options)
-      # add a 'data' event listener for the spawn instance
       stdout = ""
       stderr = ""
-      cmd.stdout.on('data', (data) -> stdout += data )
-      cmd.stderr.on('data', (data) -> stderr += data )
-      # when the spawn child process exits,
-      # check if there were any errors and
-      # close the writeable stream
+
+      cmd.stdout.on('data', (data) ->
+        stdout += data
+      )
+      cmd.stderr.on('data', (data) ->
+        stderr += data
+      )
       cmd.on('close', (returnCode) =>
         @debug('spawn done', returnCode, stderr, stdout)
         resolve({returnCode, stdout, stderr})
@@ -321,6 +328,8 @@ module.exports = class Beautifier
         @debug('error', err)
         reject(err)
       )
+
+      onStdin cmd.stdin if onStdin
     )
 
   ###
--- a/src/beautifiers/beautifier.coffee
+++ b/src/beautifiers/beautifier.coffee
@@ -313,7 +313,7 @@ module.exports = class Beautifier
       # when the spawn child process exits,
       # check if there were any errors and
       # close the writeable stream
-      cmd.on('exit', (returnCode) =>
+      cmd.on('close', (returnCode) =>
         @debug('spawn done', returnCode, stderr, stdout)
         resolve({returnCode, stdout, stderr})
       )
--- a/src/beautifiers/rustfmt.coffee
+++ b/src/beautifiers/rustfmt.coffee
@@ -4,9 +4,11 @@ Requires https://github.com/nrc/rustfmt
 
 "use strict"
 Beautifier = require('./beautifier')
+path = require('path')
+
+versionCheckState = false
 
 module.exports = class Rustfmt extends Beautifier
-
   name: "rustfmt"
 
   options: {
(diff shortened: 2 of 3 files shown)


sidred:
Isn't the config-path option still required? Will this pick up the rustfmt.toml definitions in the project root.

lhecker:
No, @sidred, it's not required anymore. I've added [code](https://github.com/Glavin001/atom-beautify/pull/975/files#diff-395fce75c484501462367319512c3b29R233) to allow setting the working directory (cwd) of commands run by `Beautifier::run()`. Thus I simply spawn `rustfmt` with a [`cwd`](https://github.com/Glavin001/atom-beautify/pull/975/files#diff-de9ffa3fe2c9211819af69fccdb87145R50) of the directory the to-be-beautified file is in.

lhecker:
You guys can try this out by simply copying the 2 changed files into the `$HOME/.atom/packages/atom-beautify/src/beautifiers` folder (same on windows). Atom will pick it up on the next restart.

cheercroaker:
@lhecker Thanks! Until this is integrated in the next version of atom-beautify, applying your modifications allows module syntax to be handled flawlessly (no error with "Beautify on save" on my configuration) 👍 

OS: `Arch Linux`
Version of Atom: `1.7.3`
Version of atom-beautify: `0.29.7`
Version of rustfmt: `0.5.0`

cheercroaker:
Using these modifications I have found that from time to time, with the "Beautify On Save" setting activated, it empties the editor buffer on save. This bug does not seem to occur without the modifications.

Here is a scenario to reproduce it "quite deterministically":

Step 1. Create a new Rust binary project:
`$ cargo new hello_world --bin`

Step 2. Add the `hello_world` project folder to your Atom session
("File -> Add project folder..." and select the "hello_world" folder corresponding to the project).

Step 3. Open the existing `src/main.rs` file and type this specific sequence right after (in this sequence description, keystrokes are separated with the `+` sign and entered text is between double quotes):

`down_arrow+down_arrow+down_arrow+"fn"+space+"hey("+right_arrow+space+"{"+enter+down_arrow+enter+"fn"+space+"hi("+right_arrow+space+"{"+enter`

With Soft Tabs activated and a Tab Length of 4 spaces you should end up at position 8:5 (thus inside the `hi` function); the `src/main.rs` file would contain this:
```rust
fn main() {
    println!("Hello, world!");
}
fn hey() {
    
}
fn hi() {
    
}
```
Step 4. Save the file (with the "Beautify On Save" setting activated) by using `Ctrl-S` (or whatever your "Core: Save" keyboard shortcut is; it seems important to use the keyboard only to reproduce the bug this way). Code gets formatted and looks like this (OK as expected):

```rust
fn main() {
    println!("Hello, world!");
}
fn hey() {}
fn hi() {}
```
Step 5. Go to line 4 by typing the up arrow key twice (with the Ctrl-G/Go to line shortcut the bug does not seem as easily reproducible) and delete it with `Ctrl-Shift-K` (or your "Editor: Delete Line" shortcut).

Step 6. Save with the keyboard shortcut: the editor buffer becomes empty! (if not, save again 5 to 10 times to get it)

An easy workaround is to undo with Ctrl-Z (or your Core: Undo shortcut) then save the file again to get the file formatted correctly and saved.
@Glavin001: for me the situation is still better than without the modifications from @lhecker (so thanks again) as it allows to use Rust's module syntax while formatting files with rustfmt from Atom (I was stuck with an error otherwise and could not use the "Beautify On Save" feature as the formatting error was preventing saving; now I can move forward). I guess other users would not bear the "emptying buffer surprise" though 😉  

_Note 1_: if you use the mouse to navigate into the file at step 3/5 or to save the file, or even to select text, you might get the bug sooner (i.e. without steps 5&6)... but not necessarily every time...

_Note 2_: if you use another sequence to get the file as expected in step 4, you might not get the bug directly; but it seems that by saving multiple times then (between 5 to 10 times) allows to finally get the bug.

_Note 3_: without the "Beautify On Save" setting, the bug is apparently very less visible but still exists as well; it can be seen by repeating the "Beautify Editor" action multiple times.

Hope that helps to narrow down the circumstances of the bug, in order to get a perfect behavior of the beautifier with Rust.

lhecker:
Sorry! I ignored this issue for a couple days, because I'm quite busy right now. I'll look into the issue mentioned by @cheercroaker early next week though.

Glavin001:
Let me know when you're ready for merging 😃 . Thanks!

lhecker:
@Glavin001 If you can please wait a bit more if you can… My workload has been increasing ever since around the time I created this PR.

Glavin001:
Not a problem. Quality takes time, I understand. Let me know whenever you're ready. No rush.

lhecker:
@Glavin001 @cheercroaker Today I started working on this again. I improved my code severely and fixed a lot of bugs. And I found the issue of the disappearing editor contents… the `Beautifier::run` method spuriously fails and returns an empty string (i.e. without throwing any errors)! I'll try to investigate into this…

lhecker:
I found the issue in `Beautifier#spawn()`. Should I open a new issue for this @Glavin001?

Glavin001:
> I found the issue in Beautifier::spawn. Should I open a new issue for this @Glavin001 or do you mind if I bundle those changes right here?

What's the issue? It could depend, however you can bundle it in here. Try to make the commits separate at least, if you can, in case we do need to separate things. Thanks!

lhecker:
@Glavin001 I (force-)pushed my new commits. I wouldn't say that it's beautiful code, but I'm not very familiar with coffescript as I said.

@cheercroaker The issue is now resolved.

BTW: This project contains a lot of duplicate and outdated code. Creating an issue for each seperate cleaned up file/module would probably be too annoying though. But I still wouldn't mind cleaning it up whenever I have time again. And after using TypeScript for most of my projects I have to say that I do really miss having a type checker. Would be cool if atom-beautify could adopt one sometime in the future too...

Glavin001:
> This project contains a lot of duplicate and outdated code.

Anything in particular you'd like to mention so I could take a closer look? 
Thanks.

lhecker:
Well I feel more like the overall code quality could simply be higher by going through all files and adopting a consistent coding style and improving modularity.

For instance most code is really packed tightly without much white space to seperate blocks which do different things, making it hard to read. You'll also find code which uses `"`, while some is using `'`. There is also stuff like [this](https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/clang-format.coffee#L23-L36). Why is not using `fs.writeFile`? That's 1 LOC instead of a dozen. It should also be part of the abstract Beautifier class and not of that specific subclass. I mean... [here](https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/beautifier.coffee#L57-L71). What's this? The same code, copy/pasted over? Nononono... 😢

You should also make all member methods of the `Beautifer` class `static`, which do not directly access or modify the state. This should improve performance and memory usage by quite a margin.
Also [this](https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/index.coffee#L97-L100) should be lazy loaded, which should again severely improve performance & memory usage. And I think this is just scratching the surface.

I really don't want this comment to sound like an insult or something... I do think though that there is quite a margin for improvements here, of which we all can benefit. 😊

Glavin001:
> Well I feel more like the overall code quality could simply be higher by going through all files and adopting a consistent coding style and improving modularity.

[CoffeeScript Linting](https://github.com/Glavin001/atom-beautify/blob/master/coffeelint.json) is currently in place to enforce coding style. It could become more strict.
I plan to switch to ES6 anyway, so this is less of a concern right now.

The `Beautifier` class was developed to make adding support for new beautifiers more modular. I think this has been serving its purpose, with the exception of some more verbose implementations of beautifiers from Pull Requests, which can be cleaned up by anyone open to submitting another Pull Request.

> There is also stuff like this. Why is not using fs.writeFile? That's 1 LOC instead of a dozen. It should also be part of the abstract Beautifier class and not of that specific subclass. I mean... here. What's this? The same code, copy/pasted over? Nononono... 😢

I did not write that beautifier, however I also did not put too much time into vetting the Pull Requests either. 
I added methods in `beautifier`, such as [`tempFile`](https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/beautifier.coffee#L57-L71), such that they could be reused, however they often are missed by devs submitting Pull Requests. I opt to get things merged in quicker than to have a lot of back and forth, knowing that improvements can be just as easily made via another Pull Request while users are already benefiting from the new language and/or beautifier support.

> You should also make all member methods of the Beautifer class static, which do not directly access or modify the state.

Good idea.

> Also this should be lazy loaded, which should again severely improve performance & memory usage.

The entire Atom-Beautify package is lazy loaded by [`activationCommands`](https://github.com/Glavin001/atom-beautify/blob/master/package.json#L134-L146). Once Atom-Beautify is loaded, it needs to know all of the beautifiers and languages that are supported such that it can pick the appropriate one to handle beautification. Essentially, it would be called in the constructor or immediately afterwards, regardless.

> I really don't want this comment to sound like an insult or something... I do think though that there is quite a margin for improvements here, of which we all can benefit. 😊

Overall, good points and I absolutely agree. I know I definitely do not have the time to keep the code quality up to where I would prefer it to be. My top priority is providing functionality to Atom-Beautify users and I am always open to Pull Requests from the community. If you feel you have the time to clean all of this up, then I look forward to reviewing your Pull Request. Thanks!

lhecker:
So let's say I have a buch of improvements which do not directly relate to each other, except for making the code quality better. You wouldn't mind if I create a single PR for all of them then right?

Glavin001:
Yup, that'll wrok. Make that a separate Pull Request after this one for improved rustfmt support.
Thanks :+1:.

lhecker:
@Glavin001 A couple things:

* Is it possible for you to merge this soon? I'd like for this to be merged before I send other PRs, to reduce the amount of conflicts I cause.
* Why are the failing CI tests not fixed? Is there a specific reason or can I simply fix the expected output to match the beautified one to silence the errors? (In case I can manually verify that the beautifier works as intended.)
* I'd like to highlight [this](https://github.com/Glavin001/atom-beautify/pull/975/commits/8faa197bca11152d89ff9fc9e3639460aae445ca) commit of mine if you missed it. It might fix quite a lot of other subtile errors where the editor content was spuriously going missing.

src/beautifiers/rustfmt.coffee
```diff
@@ -4,34 +4,50 @@ Requires https://github.com/nrc/rustfmt
...
-
-    # get file path which is the search path for rustfmt.toml as
-    # the beautifier runs rustfmt in a tmp directory.
-    # This will pick up any rustfmt.toml defined in the crate root
-
     editor = atom.workspace.getActivePaneItem()
-    file = editor?.buffer.file
-    filePath = file?.path
+    buffer = editor.getBuffer?()
+    filePath = buffer.getPath?()
```
Glavin001: `filePath` is now accessible with https://github.com/Glavin001/atom-beautify/pull/956

Example usage:

```coffeescript
beautify: (text, language, options, context) ->
  filePath = context.filePath
```

See https://github.com/Glavin001/atom-beautify/blob/66be231b953a6efd496a1c4f56a33fc730c0b502/src/beautifiers/marko-beautifier.coffee#L26 and https://github.com/Glavin001/atom-beautify/blob/66be231b953a6efd496a1c4f56a33fc730c0b502/docs/add-languages-and-beautifiers.md#how-to-add-a-beautifier-for-a-language for more details.

src/beautifiers/beautifier.coffee
```diff
@@ -248,79 +248,88 @@ module.exports = class Beautifier
...
-      # close the writeable stream
-      cmd.on('exit', (returnCode) =>
+
+      cmd.stdout.on('data', (data) ->
+        stdout += data
+      )
+      cmd.stderr.on('data', (data) ->
+        stderr += data
+      )
+      cmd.on('close', (returnCode) =>
```
Glavin001: Glad to hear changing `exit` to `close` fixes some issues!

Glavin001:
Hey. Sorry I've been studying for final exam tomorrow.

If this PR is all good to go I will review and merge now.

lhecker:
@Glavin001 I made the change for `context.filePath`. IMHO it's good to go. The only downside of this patch btw is that it's not compatible with older (> 1.5 months) versions of `rustfmt`. You have to upgrade the the installed version for the beautifier to work again (it correctly tells the user to upgrade `rustfmt` though as you probably already saw).

Glavin001:
Awesome work! Thanks for contributing! I'll take a look into these other failing tests this weekend.

I look forward to seeing a Pull Request with more of your improvements 😃 !

cheercroaker:
@lhecker Great! I cherry picked [this commit](https://github.com/Glavin001/atom-beautify/pull/975/commits/8faa197bca11152d89ff9fc9e3639460aae445ca) yesterday (before it was merged) and tested it for an hour while playing with a Rust incubating project: no more "random buffer swiping" so far, "Beautify on save" activated 👍
I love surgical fixes like these! 🔬

@Glavin001 Do you still plan to release it in v0.30.0 of atom-beautify or can we expect to get it in the upcoming (0.29.9?) version? (I am currently using v0.29.7 with the initial commit of this PR plus the magic fix: I prefer to keep my working configuration and wait for this to be included in a release, preventing myself to click on "update" until then 😊)

Cheers!

Glavin001:
@cheercroaker : There are some failing tests as a result of merged Pull Requests. I'm cleaning things up on my own right now. Will be publishing soon 😃 !

Glavin001:
Published to v0.29.9

PR #1025, opened 2016-06-20
Added "h" as a extension to the C language

WhoBrokeTheBuild:
This allows beautification of h files when you're doing C development

No

### Any other comments?
This might break the cpp language since it currently uses the "h" extension, I'm not sure

### Checklist

- [N/A] Added examples for testing to [examples/ directory](examples/)
- [X] Travis CI passes (Mac support)
- [X] AppVeyor passes (Windows support)

### DIFF
--- a/src/languages/c.coffee
+++ b/src/languages/c.coffee
@@ -15,6 +15,7 @@ module.exports = {
   Supported extensions
   ###
   extensions: [
+    "h"
     "c"
     "cl"
   ]


MichaelHancock:
Hmmm, yes the problem here could be that c++ uses .h files... I'm not 100% sure what the best course of action would be. However I will say that the nature of both languages is similar enough that conceivably the same beautification could be applied..?

Glavin001:
The editor grammar will take precedence so as long as you are using [`language-c`](https://github.com/atom/language-c/blob/master/grammars/c.cson#L7) which has grammar named `C` Atom-Beautify will use the appropriate language with matching grammar https://github.com/WhoBrokeTheBuild/atom-beautify/blob/4b0cfc41a3a5748ee2eeeeb73fed6ec79bf05ea7/src/languages/c.coffee#L10
If you want to force `C++` then switch grammar to `C++`: https://github.com/atom/language-c/blob/master/grammars/c%2B%2B.cson#L22

So this change is fine 👍.

Glavin001:
Thank you for contributing!

Glavin001:
Published to v0.29.9

PR #1062, opened 2016-07-11
check for env.path case-insensitive in windows

### LINKED ISSUE #1054: Could not find 'X'. The program may not be installed. on Windows 10
I was having an issue where every command that I did not explicitly set the path to was throwing an error.

`Could not find 'X'. The program may not be installed.`

I tracked the bug down to [/src/beautifiers/beautifier.coffee#L171](https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/beautifier.coffee#L171)

```javascript
        options.path ?= env.PATH
```

Turns out my `PATH` variable was set to `Path` so `env.PATH` was undefined

I have no idea how the variable name got changed but when I changed it back and restarted Atom everything worked fine.

Just throwing this out there for anyone else who may have this issue.

UziTech:
fixes #1054 and probably lots of other issues

### DIFF
--- a/src/beautifiers/beautifier.coffee
+++ b/src/beautifiers/beautifier.coffee
@@ -170,6 +170,14 @@ module.exports = class Beautifier
       new Promise((resolve, reject) =>
         options.path ?= env.PATH
         if @isWindows
+          # Environment variables are case-insensitive in windows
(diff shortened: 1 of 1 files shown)


Glavin001:
Thanks for contributing!

PR #1119, opened 2016-08-08
Fixes #722, adding language vue.js

### LINKED ISSUE #722: Add Language Support for Vue component files.
Please see http://vuejs.github.io/vue-loader/start/spec.html for a spec on vue files.

Basic premise is its an XML/HTML file that can contain Javascript, Coffee script, Less, Sass or Stylus.
For now I can make use of linking components to source files, but would be nice to have support for Vue files themselves.

prettydiff:
Adding Vue.js in support of issue #722

### DIFF
--- a/src/languages/vue.coffee
+++ b/src/languages/vue.coffee
@@ -1,0 +1,23 @@
+﻿module.exports = {
+
+  name: "Vue"
+  namespace: "vue"
+  fallback: ['html','vue']
+
+  ###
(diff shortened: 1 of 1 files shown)


prettydiff:
Would adding Vue as an independent grammar like this allow support for the options provided to HTML?  If not I will need to amend this change.

prettydiff:
Closing as #1136 is the same but substantially more precise.

PR #1135, opened 2016-08-14
Change path to uncrustify config file to be relative to the project rather than the open file.

StevenIsaacs:

When the configuration for the `uncrustify` config file path has no path
`atom-beautify` had the habit of pre-pending the path to the
file open in the editor pane. This meant that in order to use a
project specific configuration every directory would need to contain
a config file -- unacceptable. This changes the scheme to pre-pend
the path to the project when no path is specified making it possible
to have a single configuration file for an entire project and also
being able to maintain the configuration under source control.

...

### Any other comments?

...

### Checklist

### DIFF
--- a/src/beautifiers/uncrustify/index.coffee
+++ b/src/beautifiers/uncrustify/index.coffee
@@ -10,6 +10,7 @@ _ = require('lodash')
 
 module.exports = class Uncrustify extends Beautifier
   name: "Uncrustify"
+  link: "https://github.com/uncrustify/uncrustify"
   options: {
     Apex: true
     C: true
--- a/src/beautifiers/uncrustify/index.coffee
+++ b/src/beautifiers/uncrustify/index.coffee
@@ -10,7 +10,6 @@ _ = require('lodash')
 
 module.exports = class Uncrustify extends Beautifier
   name: "Uncrustify"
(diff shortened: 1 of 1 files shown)


src/beautifiers/uncrustify/index.coffee
```diff
@@ -38,16 +38,18 @@ module.exports = class Uncrustify extends Beautifier
         editor = atom.workspace.getActiveTextEditor()
         if editor?
           basePath = path.dirname(editor.getPath())
```
Glavin001: Line 40 and 43 can be removed, as `basePath` is no longer used. Thanks!

src/beautifiers/uncrustify/index.coffee
```diff
@@ -38,16 +38,18 @@ module.exports = class Uncrustify extends Beautifier
...
-          # Expand Home Directory in Config Path
-          configPath = expandHomeDir(configPath)
+          projectPath = atom.workspace.project.getPaths()[0]
```
Glavin001: There does not appear to be a good way to detect which path is currently selected. See https://discuss.atom.io/t/get-the-active-project/29309
This will have to do for now.

Glavin001:
Thank you for contributing!

Glavin001:
Published to v0.29.12

Issue #1088, opened 2016-07-24
Sass format don't work properly


raky291:
My code:
![err1](https://cloud.githubusercontent.com/assets/12045722/17082593/dfa7afe4-5141-11e6-8e63-ba4b527049d0.jpg)

After pressing Ctrl+Alt+B
![err2](https://cloud.githubusercontent.com/assets/12045722/17082594/dfdb377e-5141-11e6-8724-ccb8d727aee5.jpg)

Any idea what's going on?

**Update:**

> Pretty Diff will not support this. The original file does not contain curly braces to separate the selector from its properties and there are no semicolons to separate the properties from each other. https://github.com/Glavin001/atom-beautify/issues/981#issuecomment-223856949

Pretty Diff don't support Sass, it cannot be the default beautifier.

And when I use CSScomb I get the following error:
![err3](https://cloud.githubusercontent.com/assets/12045722/17082761/6286b168-5146-11e6-9b3a-d6ace5053d23.jpg)

prettydiff:
Pretty Diff supports the SCSS variant of SASS.

aidistan:
There are lots of differences between SCSS and SASS. The former could be handled well with `csspretty.js` (used by prettydiff) while the latter could not.  Surely we should change the default beautifier of SASS.

@raky291 I have tried your example codes with CSScomb and everything works fine.

aidistan:
@raky291 The same error was raised by CIs. I will have a look into it.

https://travis-ci.org/Glavin001/atom-beautify/jobs/153043688
https://ci.appveyor.com/project/Glavin001/atom-beautify/build/641

raky291:
Hi @aidistan the sass syntax don't use `{}` and `;` review the following lines of code:

https://travis-ci.org/Glavin001/atom-beautify/jobs/153043688#L898
https://ci.appveyor.com/project/Glavin001/atom-beautify/build/641#L501

aidistan:
@raky291

Thanks for point out. I have updated test examples for SASS: https://github.com/Glavin001/atom-beautify/pull/1141/commits/e6bdb8a0f3aa9f3d676605bfbf7b5857e527f4bd

It turns out that `CSScomb` could not format SASS codes well either. I will try to use the official cli `sass-convert` instead.

aidistan:
My PR #1141 is ready for merge now. Hope it would help : )

raky291:
Thank you so much! :smiley:

PR #1141, opened 2016-08-17
Fix Sass beautification

aidistan:

- Add a new beautifier: SassConvert
- Fix Sass beautification

Refer to #1088 for related discussions.

...

#1088

...

### Any other comments?

This PR is ready for merge.

...

### Checklist

### DIFF
--- a/src/beautifiers/sass-convert.coffee
+++ b/src/beautifiers/sass-convert.coffee
@@ -1,0 +1,20 @@
+"use strict"
+Beautifier = require('./beautifier')
+
+module.exports = class SassConvert extends Beautifier
+  name: "SassConvert"
+  link: "http://sass-lang.com/documentation/file.SASS_REFERENCE.html#syntax"
+
+  options:
+    # TODO: Add support for options
+    CSS: false
+    Sass: false
+    SCSS: false
+
+  beautify: (text, language, options, context) ->
+    lang = language.toLowerCase()
+
+    @run("sass-convert", [
+      @tempFile("input", text),
+      "--from", lang, "--to", lang
+    ])
--- a/src/languages/sass.coffee
+++ b/src/languages/sass.coffee
@@ -2,7 +2,7 @@ module.exports = {
 
   name: "Sass"
   namespace: "sass"
-  fallback: ['css', 'scss']
+  fallback: ['css']
 
   ###
   Supported Grammars
@@ -18,6 +18,6 @@ module.exports = {
     "sass"
   ]
 
-  defaultBeautifier: "Pretty Diff"
+  defaultBeautifier: "SassConvert"
 
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,6 +86,8 @@ before_install:
   - gem install rubocop
   - gem install htmlbeautifier
   - gem install puppet-lint
+  # Sass language support
+  - gem install sass
   # Python language support
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install python; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo chmod 777 -R /opt/python; fi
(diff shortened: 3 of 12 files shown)


src/languages/sass.coffee
```diff
@@ -2,7 +2,6 @@ module.exports = {
 
   name: "Sass"
   namespace: "sass"
-  fallback: ['css', 'scss']
```
aidistan: While SCSS is an extension to CSS, SASS is not.

src/beautifiers/sass-convert.coffee
```diff
@@ -0,0 +1,20 @@
...
+
+module.exports = class SassConvert extends Beautifier
+  name: "SassConvert"
+  link: "http://sass-lang.com/documentation/file.SASS_REFERENCE.html#syntax"
+
+  options:
+    # TODO: Add support for options
+    CSS: true
+    Sass: true
+    SCSS: true
```
Glavin001: Instead of `true` (supports all options) it should be `false` (supports language with no options).

Resources
- https://github.com/Glavin001/atom-beautify/blob/master/docs/add-languages-and-beautifiers.md#how-to-add-a-beautifier-for-a-language

src/languages/sass.coffee
Glavin001: The `fallback` field is for options fallback handling. For example, if the user has not explicitly set options for `sass` then Atom-Beautify can use options configured for `css` as a fallback. This allows developers to configure for `css`, `html`, `javascript`, etc and not have to worry about explicitly configuring  languages that are similar yet different such as `scss`, `sass`, `less`, `handlebars`, `jade`, `typescript`, `flow`, etc.

And so I would at least prefer to keep `fallback: ['css']`. Let me know what you think.

Glavin001:
This is great so far! Please take a look at my comments. Let's get this merged ASAP!!

src/languages/sass.coffee
aidistan: The `fallback` field is a great design. I deleted this line simply because Sass has no fallbacks. It's another way to write CSS and could be compiled into CSS. however, it's completely different from CSS (no braces, indents as scopes, etc).

I don't think any CSS beautifier could handle Sass well. Take #1088 as an example.

src/beautifiers/sass-convert.coffee
aidistan: I have changed those options to `false` already. Sorry for missing that part.

Maybe a notice for running `npm run docs` should be added into the docs, since 8c77514 updated the docs much more than what I did :smile:

aidistan:
I have replied your comments, amended one commit and added one new for document update.

Travis CI was broken due to "brew tap homebrew/homebrew-php" failing. Maybe a simple rebuild would fix it.

Glavin001:
I just pushed a new release. Please resolve merge conflicts and hopefully then the tests will all pass! Thank you!

aidistan:
I have resolved the conflicts and Travis CI was still broken due to the same reason.

aidistan:
Can we merge this PR? Or is there something else blocking it?

src/languages/sass.coffee
Glavin001: I would prefer to have `fallback: ['css']` remain (be put back) such that users that have already configured `css` will be able to immediately use `Sass` without any changes. Advanced `Sass` users will, of course, be able to edit configuration under the `sass` namespace which will take precedence over `css` namespace.

aidistan: Done and pushed.

src/languages/sass.coffee
```diff
@@ -2,7 +2,7 @@ module.exports = {
...
+  fallback: ['css']
```
aidistan: Now we have `fallback: ['css']` back.

Glavin001:
Could you update the docs again? Looks like they are out of data with the options set to `false` (support Language, no supported options). I think that's all that's left! Thanks!

aidistan:
I had thought about it and tried, but got no difference.

Glavin001:
Hmm. Try running:

```bash
npm run postinstall
npm run docs
```

I think the issue is that `npm run docs` was using an old `options.json` file. `npm run postinstall` should update it accordingly.

aidistan:
Got it. Done.

Glavin001:
Awesome! Thank you for contributing!

Glavin001:
Published to v0.29.12

qwloh:
@Glavin001 

Looking forward to Unibeautify's integration to Atom! Meanwhile, I ran into some issues while trying to set up sass beautifier in current atom-beautify. I've installed sass-convert as instructed in README but atom-beautify seems still not able to find the necessary files to beautify SASS. I've included a gif showing the issue and my setup here. Would you be able to point out what might be wrong? Thanks

sass-convert

Issue #1018, opened 2016-06-14
Weird spaces in JSX html tags


nixgadget:
### Description

I have the default options for JSX and my `.jsbeautifyrc` is as follows,

```
{
    "indent_size": 4,
    "indent_char": " ",
    "other": " ",
    "indent_level": 0,
    "indent_with_tabs": false,
    "preserve_newlines": true,
    "max_preserve_newlines": 2,
    "jslint_happy": false,
    "indent_handlebars": true,
}
```

But in the following JSX code you can see weird spaces in tags e.t.c

```
ReactDOM.render( < div > This is rendered by React < App translator = {
        translator
    }
    /></div > ,
    document.getElementById('react-container')
);
```
### Expected Results

The beautified code should have looked like:

```
ReactDOM.render(
    <div>This is rendered by React <App translator = {translator} /></div>,
    document.getElementById('react-container')
);
```
### Steps to Reproduce
1. Add code to Atom editor
2. Run command `Atom Beautify: Beautify Editor`
3. This beautified code does not look right!
### Debug

Here is a link to the `debug.md` Gist: https://gist.github.com/nixgadget/3f50059d0710498ed5f15e8235b46ca5
# Checklist
- [x] I have tried uninstalling and reinstalling Atom Beautify to ensure it installed properly
- [x] I have reloaded (or restarted) Atom to ensure it is not a caching issue
- [x] Searched for existing Atom Beautify Issues at https://github.com/Glavin001/atom-beautify/issues 
  so I know this is not a duplicate issue
- [x] Generated debugging information and added link for `debug.md` Gist to this issue

prettydiff:
@bitwiseman added

prettydiff:
I am not sure this issue is valid.  If the original source sample contained tags with spaces separating the delimiters from the tag names then the tags are invalid, ex: `< div >`.

nixgadget:
The problem is original code did not have spaces. But as soon as beautification is run it turns into that. What I would've expected is shown under the Expected Results.

nixgadget:
AFAIK out of the box atom-beautify uses `prettydiff` for this. Does that mean I need to define a `.prettydiffrc` ?

nixgadget:
If i run `prettydiff` and `js-beautify` you can see the difference between the two. It appears prettydiff is never called for `jsx` files.

```
[...]

Issue #974, opened 2016-05-16
How to configure atom-beautify package to format with babel/jsx javascript files?


damienleroux:
I'm new to Atom. And I need help on configuration. I post on [stackoverflow](https://stackoverflow.com/questions/37232310/how-to-configure-atom-beautify-package-to-format-with-babel-jsx-javascript-files) but without relevant answer, I seek for help here.

Thank you

Glavin001:
Please follow the issue template provided. More specifically, adding a link to the required `debug.md` Gist which includes debugging information that answers our most commonly asked questions: https://github.com/Glavin001/atom-beautify/blob/master/ISSUE_TEMPLATE.md#how-to-create-debugmd-gist
Thank you.

damienleroux:
### Description

atom-beautify doesn't indent jsx code in javascirpt file

### Expected Results

atom-beautify should indent jsx as expected when indenting html code. see [language-babel](https://atom.io/packages/language-babel) that indent only jsx code ( not the jaascript around)

### Steps to Reproduce

1. Copy code from [this javascript file](https://github.com/damienleroux/react-grid-layout-builder/blob/master/src/bootstrapEditor.js): it's common javascript + jsx code
2. Format the file:all html tag (that is jsx code) doesn't indent as expected

Thank you for any advices :)

Glavin001:
Hey @damienleroux. Thanks for elaborating 😃 .
I still need the `debug.md` Gist:
> More specifically, adding a link to the required `debug.md` Gist which includes debugging information that answers our most commonly asked questions: https://github.com/Glavin001/atom-beautify/blob/master/ISSUE_TEMPLATE.md#how-to-create-debugmd-gist
Thanks!

rr20eid10:
In Atom, I tried to install and use this plugin, but after installation, I got an underscore.js module error.  I clicked the button to create an issue, but nothing happened, so I couldn't create an issue that way.  I'm doing a tutorial online and this plugin was recommended, but when looking at description, it doesn't list JSX files specifically.  Will JSX support be added to this plugin?

Glavin001:
@rr20eid10 JSX should be supported now by JS-Beautify and Pretty Diff beautifiers.

Issue #888, opened 2016-04-03
`.es6` Support


willin:
### Description

```
Atom Beautify could not find a supported beautifier for this file
Atom Beautify could not determine a supported beautifier to handle this file with grammar "Babel ES6 JavaScript" and extension "es6". If you would like to request support for this file and its language, please create an issue for Atom Beautify at https://github.com/Glavin001/atom-beautify/issues
```

### Expected Results

Just Like `.js` Files

Glavin001:
I would review and merge a Pull Request changing https://github.com/Glavin001/atom-beautify/blob/master/src/languages/javascript.coffee#L25

PR #1186, opened 2016-09-10
Add JavaScript and JSX language support for language-babel grammar.

gandm:

Adds language support in JavaScript and JSX for [language-babel](https://atom.io/packages/language-babel).

The following issues mention language-babel (`Babel ES6 JavaScript`) https://github.com/Glavin001/atom-beautify/issues/974, https://github.com/Glavin001/atom-beautify/issues/1018, https://github.com/Glavin001/atom-beautify/issues/888, https://github.com/Glavin001/atom-beautify/issues/144

### Checklist

### DIFF
--- a/src/languages/jsx.coffee
+++ b/src/languages/jsx.coffee
@@ -10,6 +10,7 @@ module.exports = {
   grammars: [
     "JSX"
     "JavaScript (JSX)"
+    "Babel ES6 JavaScript"
   ]
 
   ###
--- a/src/languages/javascript.coffee
+++ b/src/languages/javascript.coffee
@@ -16,6 +16,7 @@ module.exports = {
   ###
   grammars: [
     "JavaScript"
+    "Babel ES6 JavaScript"
   ]
 
   ###
--- a/src/languages/javascript.coffee
+++ b/src/languages/javascript.coffee
@@ -16,7 +16,6 @@ module.exports = {
   ###
   grammars: [
     "JavaScript"
-    "Babel ES6 JavaScript"
   ]
 
   ###


src/languages/jsx.coffee
```diff
@@ -10,6 +10,7 @@ module.exports = {
   grammars: [
     "JSX"
     "JavaScript (JSX)"
+    "Babel ES6 JavaScript"
```
Glavin001: My only concern is that grammars have the highest priority, over file extension. So by having `Babel ES6 JavaScript` on both files it may incorrectly pick the language. However, Pretty Diff does support both JavaScript and JSX well, I'm not sure about JS-Beautify. So the beautifier could think an ES6 JSX file is actually an ES6 JavaScript file and consequently fail to beautify correctly.
Ideally there would be `Babel ES6 JavaScript (JSX)` (or similar) to indicate a difference in grammar and language between JSX and JavaScript.

gandm: Oh I didn't know about the priority. It would probably be better if I only specified the language inside the jsx hive. I guess in that way it would also pick up pretty-diff, which [now supports flow](https://github.com/prettydiff/prettydiff/issues/362), for all JavaScript files. If that is correct I will redo the PR?

 In my experience a lot of files, probably the majority, that have jsx content use a suffix of js. For example, [Facebook's draft.js editor](https://facebook.github.io/draft-js/) uses .js filetypes with jsx inside [here](https://facebook.github.io/draft-js/). I guess assuming a filetype is associated with content is a bit void. E.g. I also support typed flow which uses .js, and I support notion such as @decorators, class property shorthand, bind notation,..... which aren't in the JavaScript spec.

Glavin001: I think this is good. Thanks!

PR #1187, opened 2016-09-11
Add support for Nunjucks using PrettyDiff

rtablada:
Looking at the PrettyDiff beautifier, it seems that the support for nunjucks templates already exists.
But, this PR adds the required config to hook up atom-beatuify to recognize the Nunjucks grammar.


src/beautifiers/prettydiff.coffee
```diff
@@ -48,6 +48,7 @@ module.exports = class PrettyDiff extends Beautifier
     EJS: true
     HTML: true
     Handlebars: true
+    Nunjucks: true
```
Glavin001: Pretty Diff is using `auto` by default, which will likely work. However, there should also be a `case` below for handling `Nunjucks` and mapping to the appropriate Pretty Diff language, such as `html`.

prettydiff:
I just looked at Nunjucks.  The vocabulary and syntax appears to be identical to Twig HTML templates, which Pretty Diff supports.

Glavin001:
Thanks, @prettydiff !

Glavin001:
Published to v0.29.12

Glavin001:
Thank you @rtablada for contributing!

PR #1430, opened 2016-12-29
Add JSX option to to JS Beautify

NodeJoSe:

I would love to use the **language-babel** syntax package and **JS Beautify** together for my **.js** files, this will make it possible!


Glavin001:
Published to v0.29.15

Issue #136, opened 2014-11-05
EJS Beautification support


karloluis:
It might be hard to deal with given that it will work as a html an html and js thing running over eachother but hey. It'd be awesome and you'd be one of the few if not the most complete out there yet.

Glavin001:
@prettydiff is this on your roadmap?

prettydiff:
Yes, assign this issue to me.  I believe EJS is:
http://www.embeddedjs.com/

I was looking at something similar this morning, but with different delimiters: {@ and @} instead of <% and %>, but all the logic appears the same. The delimiters not a concern (even the angle braces).  The big challenge I am thinking through is:

    <% } else { %>

It is challenging because that is a closing and opening of a block consolidated to a single markup token. HTML/XML does not have such a concept, because markup tokens either close or open blocks or neither.  I will think through this.

The worst case scenario if I cannot find an answer is that children of such blocks do not receive an additional level of indentation.

Glavin001:
I think everything between `<%` and `%>` is JavaScript, so could you have it use two separate beautifiers? HTML for the main code and omit anything inside of the `<% %>`s and then process those next with JavaScript? I think trying to handle it as one language may become complicated.

Glavin001:
@prettydiff has added support so this is not high priority.

Glavin001:
@karloluis can you please provide an example file that I can use for my tests. Once I have passing tests I will be publishing. Thank you.

Glavin001:
Will be supported, thanks to @prettydiff, when #282 is released.

Glavin001:
Published to v0.25.0

This is a very big release with lots of internal changes and (hopefully) no breaking changes. Please let me know if you are experiencing any unusual behaviour / breaking changes after updating. Thank you.

adammeola01:
gives me a pinwheel dialogue that doesn't terminate...

prettydiff:
@adammeola01 Would you mind creating a new issue for that and following the template with the required information?  That way we can triage and diagnose it properly.

mywarr:
@prettydiff 
had the same problem these days,
the problem only occurs on js-beautify; pretty diff is all ok.

procedure to reproduce this:
1. open an empty file
2. select grammer as HTML (Angular)
3. ctrl + option + b = endless in progress wheel spinning.

it got me a endless pinwheel even when i run the command  "Atom Beautify: Help Debug Editor" .
so i ran debug editor on pretty diff, hope it helps.

[...]

PR #1432, opened 2016-12-30
Fix JS beautify on JSX and EJS

NodeJoSe:

When i tried to beautify **EJS** or **JSX** with JS Beautify, it wont work, because i forgot to add **EJS** and **JSX** on the **Switch** when i added it to **JS Beautify** options. This will fix the issue.

### DIFF
--- a/src/beautifiers/js-beautify.coffee
+++ b/src/beautifiers/js-beautify.coffee
@@ -33,7 +33,7 @@ module.exports = class JSBeautify extends Beautifier
     return new @Promise((resolve, reject) =>
       try
         switch language
-          when "JSON", "JavaScript"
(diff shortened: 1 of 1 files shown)


Glavin001:
Published to v0.29.17

I did not know the severity of this bug! It has caused HUGE problems for many, many users! See #1458, #1471, #1483, #136, #1452. I was waiting to publish a new release after some more Pull Requests have been merged, however this completely broke JS-Beautify. Thank you for fixing it!

I have added https://github.com/Glavin001/atom-beautify/commit/46d49750e4b32baf536fafeb2e401dfedd630623 to now throw an error so this cannot ever occur again!

NodeJoSe:
Nice, i'm glad it helped!

Issue #1457, opened 2017-01-10
atom-beautify always use 'LF' as default even though the config is 'CRLF' (jsx / Pretty Diff)


deforder:
# Description

   As the title said, When I used beautify function on .jsx file (Pretty Diff), the result always used '\n' as eol even though I have set line-ending-selector as 'CRLF' .

   I'm not sure if this is some kinds of user-error which I need to do some other configuration or not. However, after I spent my time looking into the source codes and documents, I found out that the lastest version of atom-beautify haven't pass CRLF option to PrettyDiff which make the latter use 'lf' as default value.

> src\beautifiers\prettydiff.coffee

[Original Code](https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/prettydiff.coffee)

The lastest version of this file seem to missed some part of code used for checking default eol so after I modified it a bit,it turn out working perfectly as it should be.

# Modified Code (Part 1)
```  getDefaultLineEnding= ->
    switch atom.config.get('line-ending-selector.defaultLineEnding')
      when 'LF'
        return false
      when 'CRLF'
        return true
      when 'OS Default'
        return if process.platform is 'win32' then true else false
      else
        return false
```

# Modified Code (Part 2)

```     
       args =
        source: text
        lang: lang
        mode: "beautify"
        crlf: getDefaultLineEnding()
      # Merge args intos options
      _.merge(options, args)

```

# Full Source Code
```
"use strict"
Beautifier = require('./beautifier')

module.exports = class PrettyDiff extends Beautifier
  name: "Pretty Diff"
  link: "https://github.com/prettydiff/prettydiff"
  options: {
    # Apply these options first / globally, for all languages
    _:
      inchar: ["indent_with_tabs", "indent_char", (indent_with_tabs, indent_char) ->
        if (indent_with_tabs is true) then \
          "\t" else indent_char
      ]
      insize: ["indent_with_tabs", "indent_size", (indent_with_tabs, indent_size) ->
        if (indent_with_tabs is true) then \
          1 else indent_size
      ]
      objsort: (objsort) ->
        objsort or false
      preserve: ['preserve_newlines', (preserve_newlines) ->
        if (preserve_newlines is true ) then \
          "all" else "none"
      ]
      cssinsertlines: "newline_between_rules"
      comments: ["indent_comments", (indent_comments) ->
        if (indent_comments is false) then \
          "noindent" else "indent"
      ]
      force: "force_indentation"
[...]

PR #1461, opened 2017-01-11
Updated prettydiff.coffee to support EOL Detection

deforder:

As stated in [Issue 1457](https://github.com/Glavin001/atom-beautify/issues/1457), Pretty Diff required "**crlf**" parameter to set EOL type.

[Issue 1457](https://github.com/Glavin001/atom-beautify/issues/1457)

Previously,this error had been reported by **affanshahid** in [Issue 1009](https://github.com/Glavin001/atom-beautify/issues/1009)

### Any other comments?

This error used the same logic in [Issue 707](https://github.com/Glavin001/atom-beautify/pull/941) with some minor changes.

### DIFF
--- a/src/beautifiers/prettydiff.coffee
+++ b/src/beautifiers/prettydiff.coffee
@@ -117,6 +117,7 @@ module.exports = class PrettyDiff extends Beautifier
         source: text
         lang: lang
         mode: "beautify"
+        crlf: getDefaultLineEnding() # solved issue 1457
 
       # Merge args intos options
       _.merge(options, args)
@@ -130,3 +131,23 @@ module.exports = class PrettyDiff extends Beautifier
       resolve(result)
 
     )
+
+  # Retrieves the default line ending based upon the Atom configuration
+  #  `line-ending-selector.defaultLineEnding`. If the Atom configuration
+  #  indicates "OS Default", the `process.platform` is queried, returning
+  #  CRLF for Windows systems and LF for all other systems.
+  # Code modified from atom/line-ending-selector
+  # returns: The correct line-ending character sequence based upon the Atom
+  #  configuration, or `null` if the Atom line ending configuration was not
+  #  recognized.
+  # see: https://github.com/atom/line-ending-selector/blob/master/lib/main.js
+  getDefaultLineEnding= ->
+    switch atom.config.get('line-ending-selector.defaultLineEnding')
+      when 'LF'
+        return false
+      when 'CRLF'
+        return true
+      when 'OS Default'
+        return if process.platform is 'win32' then true else false
+      else
+        return false


src/beautifiers/prettydiff.coffee
```diff
@@ -117,6 +117,7 @@ module.exports = class PrettyDiff extends Beautifier
         source: text
         lang: lang
         mode: "beautify"
+        crlf: getDefaultLineEnding() # solved issue 1457
```
Glavin001: Although I see now another commit also incorrect did this: https://github.com/Glavin001/atom-beautify/blob/c3e554d9c48ac6880bed8bd9bf3bd8407b855438/src/beautifiers/js-beautify.coffee#L32

src/beautifiers/prettydiff.coffee
```diff
@@ -130,3 +131,23 @@ module.exports = class PrettyDiff extends Beautifier
...
+  # Retrieves the default line ending based upon the Atom configuration
+  #  `line-ending-selector.defaultLineEnding`. If the Atom configuration
+  #  indicates "OS Default", the `process.platform` is queried, returning
+  #  CRLF for Windows systems and LF for all other systems.
+  # Code modified from atom/line-ending-selector
+  # returns: The correct line-ending character sequence based upon the Atom
+  #  configuration, or `null` if the Atom line ending configuration was not
+  #  recognized.
+  # see: https://github.com/atom/line-ending-selector/blob/master/lib/main.js
+  getDefaultLineEnding= ->
```
Glavin001: - [ ] Can you move this into `beautifier.coffee` along with https://github.com/Glavin001/atom-beautify/blob/c3e554d9c48ac6880bed8bd9bf3bd8407b855438/src/beautifiers/js-beautify.coffee#L62-L80 ?
Since they are used in multiple places it should be centralized within the `beautifier` class.

src/beautifiers/prettydiff.coffee
Glavin001: @prettydiff can you confirm the option is named `crlf`? That seems strange to me, instead of `eol`.

Glavin001: - [ ] Support custom `eol` from `options`. See https://github.com/Glavin001/atom-beautify/blob/c3e554d9c48ac6880bed8bd9bf3bd8407b855438/src/beautifiers/js-beautify.coffee#L32 for example.

Glavin001: - [ ] Update options documentation to reflect the `eol` option support in Pretty Diff

Glavin001: You will need to add `eol` option to the languages which should support this option, such as https://github.com/Glavin001/atom-beautify/blob/master/src/languages/javascript.coffee#L33

prettydiff: Yes, that is the correct option name and yes `eol` is a more intelligent name.  http://prettydiff.com/documentation.xhtml#crlf

deforder:
Okay, I will try to finish it within tomorrow if possible. (Today, I have to clear my work first)

Check list
- Remove comment
- Add eol option for js, jsx 
- Centralize the code

I think I will have to create dictionary for beautifier as well to support alias option name and value.

For example, Create config.json which contains alias data for specific option then make function to convert passing args.

Ex. translate
      Alias name - eol to crlf 
      Alias val - CRLF ( true ) / LF ( false )

I will submit a pull request first once I finish and then we will discuss about this later.

Glavin001:
> I think I will have to create things like dictionary as well to support alias option name and value.
>
> For example, Create config.json which contains alias data for specific option and make function to convert passing args base on this file.
>
> Ex. Map
Alias name - eol to crlf
Alias val - CRLF ( true ) / LF ( false )

I do not follow what you mean. I do not want you to accidentally do unneeded work so I wanted to make sure you know Atom-Beautify supports different types of options:
- Renaming / alias: see https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/prettydiff.coffee#L35-L38 for example
- function mapping: https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/prettydiff.coffee#L18-L19
- function mapping with rename and/or dependencies: see https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/prettydiff.coffee#L20-L23 and https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/prettydiff.coffee#L10-L17

I expect you to have something like `"crlf": "eol"` option in `Pretty Diff` and add `eol` (boolean) option to the respective languages. 

I hope that helps!

deforder:
Thanks for your guidance! 

I've almost finished this one. However, I've encountered some issue during unit test.

If I test on .js file while using js-beautifer as prefer beautifer, atom-beautifier would detect the file as jsx.

As a result, Atom will always use js config option and since .jsx isn't support by js-beautifier, PrettyDiff will be the only beautifier left for the task.

Is this intentional to make js use jsx configuration as default?

PR #1469, opened 2017-01-14
Added override EOL option for JS , JSX

deforder:

[+] Added override EOL option for JS (Supported js-beautifier and Pretty Diff)
[+] Centralized getDefaultLineEnding code and supported alias CRLF, LF parameters.
[-] Removed unnecessary comment. [#issue , #TODO in js-beautifier.coffee]

[Issue 1457](https://github.com/Glavin001/atom-beautify/issues/1457)
Previously, this error had been reported by **affanshahid** in Issue [Issue 1009](https://github.com/Glavin001/atom-beautify/issues/1009)

### Any other comments? 

- I've closed the previous pull request since the scope was changed. 
- Could @Glavin001 explain to me, how do I make EOL option only appears in Javascript and JSX configuration? (Currently, This option appears in multiple places such as ' Vue , CoffeeScript and so on.') 
- options.md file which I've commit seem to be incorrect (too many deletions). This mean I have to manually integrate new one from a generated file, is that right?
- if I have language-babel installed as Atom-package, it seems atom-beautifier always prioritize JSX over JS even though it is .js file. This will cause the selected language become JSX as well as using JSX configuration.

### Checklist


Glavin001:
> Could @Glavin001 explain to me, how do I make EOL option only appears in Javascript and JSX configuration? (Currently, This option appears in multiple places such as ' Vue , CoffeeScript and so on.')

So far what you have done looks correct. It is likely a result of using `fallback` in those languages, such as `Vue` and `CoffeeScript`, etc.

> options.md file which I've commit seem to be incorrect (too many deletions). This mean I have to manually integrate new one from a generated file, is that right?

I expect insertions, not deletions. You should be able to automatically generate the `options.md` file complete, even if you delete it first. No manual changes required, in fact manual changes would be highly discouraged.

> if I have language-babel installed as Atom-package, it seems atom-beautifier always prioritize JSX over JS even though it is .js file. This will cause the selected language become JSX as well as using JSX configuration.

I'm not yet sure what to do about this. It is rather an Atom grammar bug in the sense that JavaScript files should be recognized as JS not JSX. However, I am not sure how this is decided. After Atom determines the grammar, then Atom-Beautify uses it. Therefore, this is out of Atom-Beautify's hands and the issue lies in either the language package or Atom itself.

deforder:
> I expect insertions, not deletions. You should be able to automatically generate the options.md file complete, even if you delete it first. No manual changes required, in fact, manual changes would be highly discouraged.

I just found out that EOL option in js-beautifier seems to be the cause. Could you review on js-beautify.coffee and check if I have applied global language option correctly.

deforder:
> I expect insertions, not deletions. You should be able to automatically generate the options.md file complete, even if you delete it first. No manual changes required, in fact, manual changes would be highly discouraged.

I think I understand why the document is generated that way.

In beautifier.coffee, there is one logic which I don't fully understand yet.

``` 
    if @options._?
      **globalOptions = @options._**
      delete @options._
      # Only merge if globalOptions is an object
      if typeof globalOptions is "object"
        # Iterate over all supported languages
        for lang, options of @options
          #
          if typeof options is "boolean"
            if options is true
              **@options[lang] = globalOptions**
          else if typeof options is "object"
            @options[lang] = _.merge(globalOptions, options)
          else
            @warn("Unsupported options type #{typeof options} for language #{lang}: "+ options)
```

because of this code, if I used

```
    _:
      eol: ['end_of_line', (end_of_line) ->
        if (end_of_line == 'CRLF')
          '\r\n'
        else if (end_of_line == 'LF')
          '\n'
        else
          null
      ]
```

in js-beautifier.coffee then everything will become mess up. This because globalOptions(aka. _ ) will replace previous options object that retrieves from the setting (if that language is set to 'true' in beautifier).

This mean, I can't use ( _ ) to initiate the value but I have to put it somewhere else, right?

deforder:
> I expect insertions, not deletions. You should be able to automatically generate the options.md file complete, even if you delete it first. No manual changes required, in fact manual changes would be highly discouraged.

It seems there is something wrong with my code. If I didn't put eol under ._ in js-beautify.coffee and instead I've put it like this in **function beautify()**

```
    end_of_line : options.end_of_line
    options.eol = (end_of_line) ->
        if (end_of_line == 'CRLF')
          '\r\n'
        else if (end_of_line == 'LF')
          '\n'
        else
          @getDefaultLineEnding('\r\n','\n')
```

then everything seems to be alright. However, If I use the code as I had commit then the generated file will be broken as you saw.

I also found out something amiss, when I debugging to view option arguments that pass to prettydiff.coffee in Atom-beautify current published version. I saw that there aren't any values from setting option in **function beautify** unless you'd declared its first as a global option.

so, in order to get those options, I have to put the code like this

options: {
    ._ : 
         default_beautifier : default_beautifier
.
.
.
}

Do you have any idea why it becomes like this? Instead of merge parameters from settings, it seems to discard all setting parameter unless I've predefined every option I required as I mentioned above.

deforder:
@Glavin001 , I have changed my logic to make EOL option not to override those options in setting and update the document.

please review and let me know if there's anything need to be changed.

src/beautifiers/beautifier.coffee
```diff
@@ -97,6 +97,28 @@ module.exports = class Beautifier
...
+  #  `line-ending-selector.defaultLineEnding`. If the Atom configuration
+  #  indicates "OS Default", the `process.platform` is queried, returning
+  #  CRLF for Windows systems and LF for all other systems.
+  # Code modified from atom/line-ending-selector
+  # returns: The correct line-ending character sequence based upon the Atom
+  #  configuration, or `null` if the Atom line ending configuration was not
+  #  recognized.
+  # see: https://github.com/atom/line-ending-selector/blob/master/lib/main.js
+  getDefaultLineEnding: (crlf,lf,optionEol) ->
+    if (!optionEol || optionEol == 'System Default')
```
Glavin001: Why `System Default` instead of `OS Default`? Is this in hopes to make it clearer for users?

deforder: It shouldn't be `OS Default` since `System Default` option will change based on what user set in `Line Ending Selector` package. 

You might call it `Atom Default`, but it could consider as `OS Default` as well if users haven't touched the default setting in that package.

Glavin001: Ooh I understand. Thanks!

Let's leave it as is and see if users have questions we can improve the documentation, or rename if necessary.

Glavin001:
Thank you for contributing!

Glavin001:
Published to v0.29.20

PR #1513, opened 2017-02-13
Add ESLint beautifier

taylon:
Hey guys,

This is an implementation for the ESLint fixer for Javascript.
Let me know if there is any improvement that must be done before it gets merged. I have been using it for a couple of weeks and it's great.

Thanks,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add [elm-format](https://github.com/avh4/elm-format) beautifier for the Elm language.
 - Add [clang-format](http://clang.llvm.org/docs/ClangFormat.html) beautifier for C/C++/Obj-C languages.
 - Add [yapf](http://github.com/google/yapf) beautifier for Python.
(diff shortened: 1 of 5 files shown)


prettydiff:
You will need to get the travis.ci tests to pass before Glavion will merge this.  You will probably also have to create test units.

Have you checked to see which of the currently supported options are applicable to ESLint and applied then in the necessary configuration?

taylon:
I just fixed a couple of lint errors but the travis job still does not pass, but I believe it is nothing related to the ESLint beautifier.

Regarding the configuration, I think that ESLint doesn't make much sense without it's own configuration so I went with an approach similar to the JSCS and other beautifiers which doesn't not support any configuration.

Glavin001:
Since this is not a default beautifier for JavaScript, it will not be a breaking change to users. I don't mind merging this and letting users test this out for themselves. 

Thank you for contributing, @taylon !

Glavin001:
Published to v0.29.18

PR #1522, opened 2017-02-21
Get isort configuration from project

taylon:
Hi,

Currently the isort implementation in the python beautifiers (yapf and autopep8) does not take the project configuration into consideration because it runs on a tempfile which makes isort kind of useless if the project have specific configurations for it.

This pull request sets the path for isort to look for the configuration to the project path, if there is no configuration in the project path isort will automatically falls back to the default behavior so this change does not affect who is already using it but fix the behavior for those who have project specific configurations.

Thanks,


Glavin001:
Thank you for contributing!

Glavin001:
Published to v0.29.19

PR #1530, opened 2017-02-27
Fixed JS Beautify JSX bug by adding e4x option

guiguan:
Originally, after beautifying JSX code using JS Beautify with latest atom-beautify, following code example:

```js
import React from 'react';

export default class App extends React.Component {
  render() {
    return (
      <div style={{textAlign: 'center'}}>
        <h1>Hello World</h1>
      </div>);
  }
}
```

would become:

```js
import React from 'react';

export default class App extends React.Component {
  render() {
    return ( <
      div style = {
        {
          textAlign: 'center'
        }
      } >
      <
      h1 > Hello World < /h1> <
      /div>
    );
  }
}
```

which is expected to stay the same


Glavin001:
Looking at the Travis CI build, the JSX tests are now failing. Need these to be fixed first before I can merge. Thanks!

Glavin001:
I changed the default beautifier back to `Pretty Diff`, since I do not want to introduce a breaking change in experience, unless users report `Pretty Diff` is not performing well and `JS Beautify` is the way to go for JSX.

Thanks for contributing!

Glavin001:
Published to v0.29.18

PR #1569, opened 2017-03-28
add simple lua format base on js.

hxsf:
1. I want to format Lua, but need install perl frist? It's not cool.
so, I use js, and translate it to coffee.

2. the current perl version has some bug like following:
- it will change whitespace in string.
`local a    = "a..b...c"` will be `local a = "a.b.c"` (use `'.'` to represent `' '`)

Of course, many of issues about 'Lua' and 'format'.

### Any other comments?
yes, I am not good at coffee, so if my code is not friendly to read. just edit it.
BTW, when I use Travis CI, Lua is OK, but other 9 language fail, I do not think I'm causing it.
### Checklist


hxsf:
wait, I found some thing wrong.

`local a = -1` will be `local a =  - 1`

I will fix it.

PR #1614, opened 2017-04-20
php-cs-fixer allow risky option

JonidBendo:

Adds an option for allow-risky in php-cs-fixer so you can apply risky rules if you want to.

I am not sure.

### Any other comments?

none.

### Checklist

### DIFF
--- a/src/beautifiers/php-cs-fixer.coffee
+++ b/src/beautifiers/php-cs-fixer.coffee
@@ -24,6 +24,7 @@ module.exports = class PHPCSFixer extends Beautifier
       "fix"
       "--rules=#{options.rules}" if options.rules
       "--config=#{configFile}" if configFile
+      "--allow-risky=#{options.allowRisky}" if options.allowRisky
       "--using-cache=no"
     ]
     if version is 1
--- a/src/beautifiers/php-cs-fixer.coffee
+++ b/src/beautifiers/php-cs-fixer.coffee
@@ -24,7 +24,7 @@ module.exports = class PHPCSFixer extends Beautifier
       "fix"
       "--rules=#{options.rules}" if options.rules
       "--config=#{configFile}" if configFile
-      "--allow-risky=#{options.allowRisky}" if options.allowRisky
+      "--allow-risky=#{options.allow_risky}" if options.allow_risky
       "--using-cache=no"
     ]
     if version is 1
--- a/src/languages/php.coffee
+++ b/src/languages/php.coffee
@@ -49,7 +49,7 @@ module.exports = {
       type: 'string'
       default: "no"
       enum: ["no", "yes"]
-      description: "allow risky rules to be applied!"
+      description: "allow risky rules to be applied (PHP-CS-Fixer 2 only)"
     phpcbf_path:
       title: "PHPCBF Path"
       type: 'string'
--- a/src/languages/php.coffee
+++ b/src/languages/php.coffee
@@ -44,6 +44,12 @@ module.exports = {
       type: 'string'
       default: ""
       description: "Add rule(s). i.e. line_ending,-full_opening_tag,@PSR2 (PHP-CS-Fixer 2 only)"
+    allow_risky:
+      title: "Allow risky rules"
+      type: 'string'
+      default: "no"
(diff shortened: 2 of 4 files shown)


Glavin001:
To confirm, this option is only available on PHP-CS-Fixer v2?

JonidBendo:
@Glavin001 This is actually something that I cannot confirm 100%. It seems to be an option for version 2 only and I only know about it because cs-fixer will return an error message and apply no beatification if:
- This option is not set to yes
- Rules that are considered risky are included in the rules to apply

src/options.json
```diff
@@ -5852,6 +5852,24 @@
           "namespace": "php"
         }
       },
+      "allow_risky": {
```
Glavin001: This should be automatically generated. I do not see a change to the PHP language adding `allow_risky` option?

JonidBendo: @Glavin001 I haven't really worked with modules before, this was my first time and I did not know about the autogeneration part, but i got it now, I also committed the change to the language file.

src/languages/php.coffee
```diff
@@ -44,6 +44,12 @@ module.exports = {
       type: 'string'
       default: ""
       description: "Add rule(s). i.e. line_ending,-full_opening_tag,@PSR2 (PHP-CS-Fixer 2 only)"
+    allow_risky:
```
Glavin001: Great! Don't forget to regenerate the documentation with `npm run docs`.

See https://github.com/Glavin001/atom-beautify/blob/master/docs/add-languages-and-beautifiers.md#options-documentation for details.

src/options.json
Glavin001: No problem! Getting closer.

src/languages/php.coffee
JonidBendo: I should have probably seen this documentation from start 👍 . Well everything should be ready now.

Glavin001: Looks good!

src/languages/php.coffee
```diff
@@ -44,6 +44,12 @@ module.exports = {
...
+      title: "Allow risky rules"
+      type: 'string'
+      default: "no"
+      enum: ["no", "yes"]
+      description: "allow risky rules to be applied!"
```
Glavin001: Could you add `(PHP-CS-Fixer 2 only)`? I should have noticed this sooner, sorry about that.

src/beautifiers/php-cs-fixer.coffee
```diff
@@ -24,6 +24,7 @@ module.exports = class PHPCSFixer extends Beautifier
       "fix"
       "--rules=#{options.rules}" if options.rules
       "--config=#{configFile}" if configFile
+      "--allow-risky=#{options.allow_risky}" if options.allow_risky
```
Glavin001: So we will always pass in `allow_risky`?

```
php-cs-fixer --allow-risky=yes
php-cs-fixer --allow-risky=no
```

Glavin001:
Looking good! I would just like to add a note about `(PHP-CS-Fixer 2 Only)` because this is not known from the information provided to non-developer users. Thanks!

src/beautifiers/php-cs-fixer.coffee
JonidBendo: @Glavin001 I thought the if there is not passing this option as long as the allow_risky is not changed (not in the configuration file), after changed yes.

I know for sure that if not defined its set to no by default. So the general behaviour would be.
1. As long as it is not in the config file it will not be included (I am not sure about this)
2. If its set to no it will be included but nothing will change from the default behaviour
3. If it is set to yes then rules like symfony and so on (that are considered risky) may be applied

Glavin001: Sounds good! I think this is good as is.

> As long as it is not in the config file it will not be included (I am not sure about this)

The default value should be present, so `no`. So as long as `php-cs-fixer --allow-risky=no` is valid, then we are all good.

Glavin001:
Thank you for contributing!

Glavin001:
Published to v0.29.24

PR #1644, opened 2017-05-05
fix `undefined` when options has no description

UziTech:
When an option doesn't have a description `undefined` prefixes the `(Supported by ...)` declaration in the settings

image

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -7074,7 +7074,7 @@ Automatically beautify LaTeX files on save
 
 **Description**:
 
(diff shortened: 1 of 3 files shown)


Glavin001:
This is great, thank you for noticing and fixing!

Could you commit the fixed documentation by running `npm run docs`?

Glavin001:
Thank you for contributing!

Glavin001:
Published to v0.29.24

Issue #1194, opened 2016-09-15
PHP code not beautifying on Windows


ShreyanshLodha:
# Description

I am a windows user using your Beautify, works beautifully on all the languages but I recently started development in PHP and I downloaded PHP-CS-FIXER and also I have set the location in the settings.
It run and give no error, but it does not beautify the code
I tried using PHPCBF but nothing works.

Please help me out.

```
"atom-beautify":
    php:
      beautify_on_save: true
      cs_fixer_path: "C:\\PHP-CS-Fixer-1.12\\PHP-CS-Fixer-1.12\\php-cs-fixer"
      phpcbf_path: "C:\\PHPCBF\\PHP_CodeSniffer-master\\scripts\\phpcbf"
      fixers: "",
      level: ""
```

guzuomuse:
+1

mintjustin:
I think a fix was provided but it is broken again. Was it something about "php" command before "php-cs-fixer"?

Glavin001:
Please follow the [issue template](https://github.com/Glavin001/atom-beautify/blob/master/ISSUE_TEMPLATE.md) provided. More specifically, [adding a link to the required `debug.md` Gist](https://github.com/Glavin001/atom-beautify/blob/master/docs/troubleshooting.md#how-to-create-debugmd-gist) which includes debugging information that answers our most commonly asked questions.
Thank you.

Issue #1643, opened 2017-05-05
phpcbf 3.0 returns ERROR: option "--no-patch" not known


UziTech:
The `--no-patch` option has been removed in [phpcbf 3.0](https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.0.0a1)

There should be some way of determining which version of phpcbf the user is using like [linter-phpcs](https://github.com/AtomLinter/linter-phpcs/blob/master/lib/main.js#L26) or a config setting like the [PHP-CS-Fixer Version](https://github.com/Glavin001/atom-beautify/blob/master/src/languages/php.coffee#L30) setting

Glavin001:
https://github.com/Glavin001/atom-beautify/issues/1645 merged.

Glavin001:
Published to v0.29.24

PR #1645, opened 2017-05-05
Add PHPCBF Version setting

UziTech:

This adds a setting to PHP for the PHPCBF Version.

If the PHPCBF version is 3 than it removes the `--no-patch` flag

#1643

### Any other comments?

The setting is defaulted to 2 so it doesn't change the current behavior

### Checklist

### DIFF
--- a/src/beautifiers/phpcbf.coffee
+++ b/src/beautifiers/phpcbf.coffee
@@ -11,12 +11,10 @@ module.exports = class PHPCBF extends Beautifier
   isPreInstalled: false
 
   options: {
-    _:
-      standard: ["standard", (standard) ->
-        if (standard) then \
-          standard else "PEAR"
-      ]
-    PHP: true
+    PHP:
+      phpcbf_path: true
+      phpcbf_version: true
+      standard: true
   }
 
   beautify: (text, language, options) ->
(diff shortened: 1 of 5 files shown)


UziTech:
I think this fixes #1194 as well

src/beautifiers/phpcbf.coffee
```diff
@@ -11,12 +11,10 @@ module.exports = class PHPCBF extends Beautifier
...
-    _:
-      standard: ["standard", (standard) ->
-        if (standard) then \
-          standard else "PEAR"
-      ]
-    PHP: true
+    PHP:
+      phpcbf_path: true
+      phpcbf_version: true
+      standard: true
```
Glavin001: I think `standard` should still be the transform function from before? i.e `standard: ["standard", (standard) ->`..

src/options.json
```diff
@@ -5874,23 +5874,41 @@
         "title": "PHPCBF Path",
         "type": "string",
         "default": "",
-        "description": "Path to the `phpcbf` CLI executable (Supported by PHP-CS-Fixer)",
```
Glavin001: Good catch.

src/beautifiers/phpcbf.coffee
UziTech: If PHPCS is called without a standard the PEAR standard is used any ways. If we want to default the standard to PEAR I feel it would be better to use the default property in options

Glavin001: Sounds good. Let's set `PEAR` as default, so we are backwards compatible. Otherwise, this looks good to go!

Glavin001:
Thanks for contributing!

Glavin001:
Published to v0.29.24

Issue #1515, opened 2017-02-14
php-cs-fixer not working on win10


kdlsw:
`See https://github.com/FriendsOfPHP/PHP-CS-Fixer for program installation instructions.
Your program is properly installed if running 'where.exe php' in your CMD prompt returns an absolute path to the executable. If this does not work then you have not installed the program correctly and so Atom Beautify will not find the program. Atom Beautify requires that the program be found in your PATH environment variable. 
Note that this is not an Atom Beautify issue if beautification does not work and the above command also does not work: this is expected behaviour, since you have not properly installed your program. Please properly setup the program and search through existing Atom Beautify issues before creating a new issue. See https://github.com/Glavin001/atom-beautify/search?q=php&type=Issues for related Issues and https://github.com/Glavin001/atom-beautify/tree/master/docs for documentation. If you are still unable to resolve this issue on your own then please create a new issue and ask for help.`

This is the error output. I am not sure how to properly install this php-cs-fixer on a win10 PC, the installation guide given in the github page are all for linux. I tired to download the php-cs-fixer.phar and the archive files, set "path to php-cs-fixer" in atom to these files, and it did not work. All the researches I have done lead to linux or osx. Can someone tell me how do deal with it on win10. Thanks

Stormiix:
same here ...
```
E:\LARAGON\www\Portfolio (master){hg}
λ where.exe php
E:\LARAGON\bin\php\php-7.1.2\php.exe
E:\LARAGON\bin\php\php-7.1.1-Win32-VC14-x86\php.exe

E:\LARAGON\www\Portfolio (master){hg}
λ where.exe php-cs-fixer
C:\Users\Anas Mazouni\AppData\Roaming\Composer\vendor\bin\php-cs-fixer
C:\Users\Anas Mazouni\AppData\Roaming\Composer\vendor\bin\php-cs-fixer.bat```
Everything is setup and configured , but same error message ,
I even tried manually downloading PHP-CS-FIXER and adding its path to Atom Beautify Settings but nothing ,
`E:\LARAGON\bin\php\php-7.1.2\php-cs-fixer-v2.phar`

DenaByte:
I have a similar issue
the PHP beautify does not work and gives the above error. Also to add up to that , the where.exe does not return anything , it speaks of INFO: Could not find files for the given pattern(s).
Which is even worse :D I tried reinstalling from ATOM and from CMD, no progress though....

Still waiting for a possible solution

[...]

PR #1649, opened 2017-05-08
add .php extension to temp files

UziTech:

Adds `.php` extension to the temp file that is created for PHPCBF

### Any other comments?

PHPCBF v3.0 will filter out and files without the `.php` extension by default. Alternatively we could tell it not to filter out the files but i felt adding the `.php` extension is easier and will help reduce errors in the future.

### Checklist

### DIFF
--- a/src/beautifiers/phpcbf.coffee
+++ b/src/beautifiers/phpcbf.coffee
@@ -50,7 +50,7 @@ module.exports = class PHPCBF extends Beautifier
             phpcbfPath unless isExec
             "--no-patch" unless options.phpcbf_version is 3
             "--standard=#{options.standard}" if options.standard
-            tempFile = @tempFile("temp", text)
+            tempFile = @tempFile("temp", text, ".php")
             ], {
               ignoreReturnCode: true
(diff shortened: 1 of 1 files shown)


36864:
Could php-cs-fixer require the same fix?

UziTech:
possibly I've never used php-cs-fixer. Do you want me to include that with this pull request?

36864:
There seem to be a number of recent-ish issues complaining about php-cs-fixer not working:
https://github.com/Glavin001/atom-beautify/issues/1487
https://github.com/Glavin001/atom-beautify/issues/1515

I checked and the temp file names are the same as for phpcbf.
In trying to test your fix for php-cs-fixer I managed to bugger my entire installation and have no time to fix it right now, but I will test it as soon as possible and report back.

If anyone can test that the same fix will indeed work for php-cs-fixer, I see no reason not fix both beautifiers at once.

Glavin001:
Thank you for contributing!

Glavin001:
Published to v0.29.26

Issue #545, opened 2015-09-11
Add support for Elixir


kyledecot:

prettydiff:
Any additional information you could provide would be helpful.  I have no idea what Elixir is.

https://github.com/Glavin001/atom-beautify/wiki/Add-an-unsupported-language-using-an-existing-beautifier

jvanek:
+1

Tonkonozhenko:
+1

prettydiff:
I am guessing Elixir is this: http://elixir-lang.org/

Tonkonozhenko:
@prettydiff, exactly

nateless:
+1

efries:
+1

davidmoshal:
+1

jeromedoyle:
+ 1

agelin:
+1

thewolf:
+1

thewolf:
http://elixir-lang.org/

greentornado:
+ 1

rakirox:
+1

mintuhouse:
+1

JohnBat26:
+1

gohackfelipe:
+ 1

sergiorivas:
+1

Dephora:
+1

chrisjowen:
+1

prettydiff:
Is anybody able to recommend a Elixir beautifier?

ssijak:
+10

prettydiff:
Somebody on Hacker News suggested https://github.com/rrrene/credo/

Does anybody have experience with that tool and does it make for a good beautifier?

Glavin001:
@prettydiff: credo looks like it only works for linting. Do you see in their documentation how to force editing the file?

prettydiff:
No, I honestly did not look closely at it.  There are many people observing this issue who are familiar with Elixir applications, so I wondering if anybody else had better information.

Glavin001:
I see 20 participants for this Issue.
Everyone who is interested in this feature being supported, no more +1s are necessary. Adding support is definitely something we want to do. What you can do to help, and what is required to make this happen, is find a third-party beautifier to integrate Atom-Beautify with.

greentornado:
Can this help ? It is intellj elixir plugin 
https://github.com/KronicDeth/intellij-elixir

ykinuse:
+1

[...]

PR #1726, opened 2017-06-19
Add goimports beautifier for Go language

tanyadegurechaff:
### What does this implement/fix? Explain your changes.
This adds goimports beautifier for Go language.

### Does this close any currently open issues?

...

### Any other comments?

...

### Checklist

Check all those that are applicable and complete.

- [x] Merged with latest `master` branch
- [x] Regenerate documentation with `npm run docs`
- [x] Add change details to `CHANGELOG.md` under "Next" section
- [ ] Added examples for testing to [examples/ directory](examples/)
- [x] Travis CI passes (Mac support)
- [x] AppVeyor passes (Windows support)

PR #1738, opened 2017-06-22
Add Elixir language formatter with elixir 1.6. Fix Glavin001#545

kumekay:

Add [Elixir language](https://elixir-lang.org) formatter with [exfmt](https://github.com/lpil/exfmt)

#545 

### Checklist


src/beautifiers/exfmt.coffee
```diff
@@ -0,0 +1,21 @@
+###
+Requires https://github.com/lpil/exfmt
+###
+"use strict"
+Beautifier = require('./beautifier')
+
+module.exports = class Exfmt extends Beautifier
+  name: "exfmt"
+  link: "https://github.com/lpil/exfmt"
+  isPreInstalled: false
```
Glavin001: Same comment as https://github.com/Glavin001/atom-beautify/pull/1726#discussion_r122724679

I would expect there to be a `mix` executable.
Then instead of `args: ['--version'],` it could be `args: ['exfmt', '--version'],`, that is if `mix exfmt --version` works?

src/beautifiers/exfmt.coffee
```diff
@@ -0,0 +1,21 @@
...
+
+  options: {
+    Elixir: true
```
Glavin001: Should be `false` since no options are supported. Although, no options exist, however if in the future options were added to the language, at least this would not incorrectly report in the documentation that `exfmt` supported those new options 😉 .

.travis.yml
```diff
@@ -168,3 +168,19 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install crystal-lang; fi
   # Bash
   - pip install beautysh
+  # Elixir
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      brew install elixir;
+    fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+      wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb;
```
Glavin001: If you were feeling really ambitious and had experience with Docker, you could contribute a `mix` and/or `exfmt` Docker image to https://github.com/Unibeautify/docker-beautifiers . However, I do not require it. Just a suggestion 😄 .

src/beautifiers/exfmt.coffee
pinya: I made an update, but how can I check that my version regex actually works?
```
version: {
        args: ['help exfmt'],
        parse: (text) -> text.match(/exfmt-(\d+\.\d+\.\d+)/)[1]
      }
```

pinya: And what should I put in `cmd` if my full command is `mix exfmt`?

Glavin001: > Full command to see version is `mix help exfmt` 

Then I would expect something like:

```
cmd: 'mix',
version: {
        args: ['help', 'exfmt'],
        parse: (text) -> text.match(/exfmt-(\d+\.\d+\.\d+)/)[1]
      }
```

The full command being: `${cmd} ${version.args.join(' '}`

Then you test it by reloading Atom and trying to beautify. 
When installed, if the version does not correctly parse it will throw an error.
Try to uninstall `exfmt` and try again, see if it throws error as you expect. So you know it works when it should work, and error when it should not (i.e. not installed).

pinya:
Glavin, 
when I don't have elixir installed and `mix` command is not available error message is ok:
```
See mix installation instructions at https://github.com/Glavin001/atom-beautify#beautifiers or go to https://github.com/lpil/exfmt
Your program is properly installed if running 'which mix' in your Terminal returns an absolute path to the executable.
```
But full command to run formatter is `mix exfmt`

And if this package is not installed error is:
```
** (Mix) The task "exfmt" could not be found
```
Which is not informative.
Is it possible to catch this error and show more descriptive/standard message?

Glavin001:
I think the `additional` help property may be useful for this: https://github.com/Glavin001/atom-beautify/blob/a2d8236d67c19d857132558b01722cb0e9b559e8/src/beautifiers/executable.coffee#L299

You could add another option for Executable: https://github.com/Glavin001/atom-beautify/blob/a2d8236d67c19d857132558b01722cb0e9b559e8/src/beautifiers/executable.coffee#L15-L42

How about the following?

> See mix installation instructions at https://github.com/Glavin001/atom-beautify#beautifiers or go to https://github.com/lpil/exfmt
Your program is properly installed if running 'which mix' in your Terminal returns an absolute path to the executable.
**If Elixir's mix is installed you can confirm `exfmt` is installed by running `mix exfmt`.**

Or something along those lines. I am open to idea. `PHP-CS-Fixer` is in a similar situtation as `exfmt` and I have not settled on a way to handle this case.

pinya:
I didn't find any examples and didn't get where to put `additional` section
I've tried to add section to exec description:

```
{
        help: {
          additional: "If Elixir's mix is installed you can confirm exfmt is installed by running mix exfmt."
        }
      }
```
But it doesn't work.
I doesn't have good knowledge of CoffeeScript, but it seems that `help` variable passed to `commandNotFoundError` is overwritten here https://github.com/Glavin001/atom-beautify/blob/a2d8236d67c19d857132558b01722cb0e9b559e8/src/beautifiers/executable.coffee#L137-L141 , so it doesn't contain additional field at the right moment.

Glavin001:
You are correct. Please feel free to add `additional` optional field back in way which works for you and I will review it. You can add `additionalHelp` to https://github.com/Glavin001/atom-beautify/blob/a2d8236d67c19d857132558b01722cb0e9b559e8/src/beautifiers/executable.coffee#L15-L42 and use it at https://github.com/Glavin001/atom-beautify/blob/a2d8236d67c19d857132558b01722cb0e9b559e8/src/beautifiers/executable.coffee#L137-L141

lpil:
Hi gang! Just stumbled across this. Exfmt is still in alpha so any feedback on the tool's CLI is very much welcomed.

If you do get exfmt and atom-beautify working nicely please do let me know so we can add it to the exfmt README. :)

Cheers,
Louis

src/beautifiers/exfmt.coffee
```diff
@@ -0,0 +1,35 @@
...
+  ]
+
+  options: {
+    Elixir: false
+  }
+
+  beautify: (text, language, options) ->
+    @run("mix", [
+      'exfmt',
+      @tempFile("input", text, ".ex")
```
lpil: exfmt has a `--stdin` flag which will allow you to avoid the use of a temp file :)

pinya: @Glavin001 Glavin, I didn't find any examples using standard input, so leaved temporary file.

.travis.yml
```diff
@@ -168,3 +168,19 @@ before_install:
...
+      sudo dpkg -i erlang-solutions_1.0_all.deb;
+      sudo apt-get update;
+      sudo apt-get install erlang;
+      kiex install 1.5.1;
+      kiex use 1.5.1;
```
lpil: Should really be the master branch. exfmt doesn't support v1.5

szeck87:
Once Elixir 1.6 is out, the format will need to change to what's described here:  https://hexdocs.pm/mix/master/Mix.Tasks.Format.html#content.

kumekay:
Finally 1.6 is released, I'll make an update this weekend
https://elixir-lang.org/blog/2018/01/17/elixir-v1-6-0-released/

src/beautifiers/exfmt.coffee
```diff
@@ -0,0 +1,36 @@
...
+      cmd: "mix"
+      homepage: "https://github.com/lpil/exfmt"
+      installation: "https://github.com/lpil/exfmt"
+      additionalHelp: "If Elixir's mix is installed you can confirm exfmt is installed by running mix exfmt or install it with 'mix archive.install github lpil/exfmt tag v0.4.0'"
+      version: {
+        args: ['help','exfmt'],
+        parse: (text) -> text.match(/exfmt-(\d+\.\d+\.\d+)/)[1]
+      }
+      docker: {
+        image: "unibeautify/exfmt"
```
Glavin001: This Docker image can be changed to use Mix 1.6 which supports `format`.
https://hexdocs.pm/mix/Mix.Tasks.Format.html

/cc @szeck87

src/beautifiers/exfmt.coffee
```diff
@@ -0,0 +1,36 @@
+###
+Requires https://github.com/lpil/exfmt
+###
+"use strict"
+Beautifier = require('./beautifier')
+
+module.exports = class Exfmt extends Beautifier
```
Glavin001: Looks like `exfmt` can be switched to using `mix`'s `format` command: https://hexdocs.pm/mix/Mix.Tasks.Format.html 🎉

lpil: A good idea, I've deprecated exfmt 👍

src/beautifiers/exfmt.coffee
Glavin001: I think this Docker image has been removed. Should use official Docker image.
cc @szeck87

src/beautifiers/exfmt.coffee
```diff
@@ -0,0 +1,36 @@
...
+  name: "exfmt"
```
Glavin001: Is `exfmt` appropriate now? Maybe a better name is `Elixir Format" since it is an official formatter.

src/beautifiers/exfmt.coffee
```diff
@@ -0,0 +1,36 @@
...
+  link: "https://github.com/lpil/exfmt"
+  executables: [
+    {
+      name: "exfmt"
+      cmd: "mix"
+      homepage: "https://github.com/lpil/exfmt"
+      installation: "https://github.com/lpil/exfmt"
+      additionalHelp: "If Elixir's mix is installed you can confirm exfmt is installed by running mix exfmt or install it with 'mix archive.install github lpil/exfmt tag v0.4.0'"
```
Glavin001: Should info about `exfmt` be removed and only use the new and official `mix format` command?

lpil: Yes please. exfmt is deprecated.

src/beautifiers/exfmt.coffee
lpil: :+1:

Issue #601, opened 2015-10-11
`.tsx` support


bennylut:
I would like to suggest support for [tsx](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#new-tsx-file-extension-and-as-operator) files which are a jsx extention for typescript.

prettydiff:
For some general background on JSX-like grammars: https://github.com/Glavin001/atom-beautify/issues/608#issuecomment-149043722

pristas-peter:
+1

olegsmetanin:
+1

sqrtroot:
+1 :+1:

prettydiff:
This morning I started evaluating support for TypeScript in Pretty Diff's JS parser (jspretty.js).  If I can correctly identify and parse TypeScript then TSX support is already available, as the parser won't know the difference between TSX and JSX.

I am not a TypeScript user myself so I will have a lot of testing to do to ensure I get this right, but also negative impact test against JavaScript and ES6/React support.  Once I get a bit further I will start posting beta test links for people to examine so that bugs can be identified.

prettydiff:
Here is something to test against:  http://prettydiff.com/ignore/testlocation/index.xhtml?m=beautify

Try out some TypeScript and tsx samples at the above tool.  If you find any errors please let me know.  In the mean time I will be looking for additional TypeScript code samples to test against and perform some regression tests.

davidtme:
+1

prettydiff:
When looking at the code example at https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#1.9 what is `<t>`?

Is that a type parameter?  Does it always follow a reference?  Could it follow an end character: `]`, `)`, `}`?  Are spaces permitted between it and the thing it is describing?

prettydiff:
Bah, https://github.com/Microsoft/TypeScript/wiki/JSX#the-as-operator

> Is this code creating a JSX element with the content of bar;, or is it asserting that bar is of type foo and there is an invalid expression on line 2? To simplify cases like this, angle bracket type assertions are not available in .tsx files. As a result, in a .tsx file, the previous code would be interpreted as a JSX element, and in a .ts file it would result in an error.

As a result could you guys please pass your TSX code samples through http://prettydiff.com/?m=beautify and let me know if they are beautified appropriately?

ShMcK:
Some issues with formatting after any dot notation. An example:

```js
    return <button
[...]

PR #1763, opened 2017-07-13
Add TSX support

luania:
support .tsx files
beautify them just like .jsx

### DIFF
--- a/src/beautifiers/js-beautify.coffee
+++ b/src/beautifiers/js-beautify.coffee
@@ -31,7 +31,7 @@ module.exports = class JSBeautify extends Beautifier
     return new @Promise((resolve, reject) =>
       try
         switch language
-          when "JSON", "JavaScript", "JSX"
+          when "JSON", "JavaScript", "JSX", "TSX"
             beautifyJS = require("js-beautify")
             text = beautifyJS(text, options)
             resolve text
--- a/src/languages/jsx.coffee
+++ b/src/languages/jsx.coffee
@@ -19,6 +19,7 @@ module.exports = {
   ###
   extensions: [
     "jsx",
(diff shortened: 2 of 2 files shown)


Glavin001:
Related to https://github.com/Glavin001/atom-beautify/issues/601

src/beautifiers/js-beautify.coffee
```diff
@@ -31,7 +31,7 @@ module.exports = class JSBeautify extends Beautifier
     return new @Promise((resolve, reject) =>
       try
         switch language
-          when "JSON", "JavaScript", "JSX"
+          when "JSON", "JavaScript", "JSX", "TSX"
```
Glavin001: `TSX` is incorrect, since the `language` in this case would actually be `JSX` from `languages/jsx.coffee`.

src/languages/jsx.coffee
```diff
@@ -19,6 +19,7 @@ module.exports = {
   ###
   extensions: [
     "jsx",
+    "tsx",
```
Glavin001: I wonder if adding to `TypeScript` language would make more sense and have better support via the TypeScript Formatter:  https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/typescript-formatter.coffee#L8

Glavin001:
A few things to fix up before we can merge. Please let me know if you have any questions! Thank you for contributing!

Glavin001:
I think using https://github.com/vvakame/typescript-formatter would be more appropriate.

Glavin001:
Closing as no changes were made after request for changes. This can be reopened if someone intends to pick up where this left off. Thanks!

PR #1883, opened 2017-10-10
Add lib to gitignore, src to npmignore, package.json & tsconfig updates

szeck87:
- Updates Typescript dependency
- If using Atom and the atom-typescript package, saving .ts files in src will compile them into .js files in lib (from change in tsconfig.json)
- Ignore lib directory in git
- Ignore src directory in npm
- Add Atom Beautify to the top menu
...

...

### Any other comments?

...

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.js
 *.js.map
 src/*.d.ts
+lib/
 
 ### https://raw.github.com/github/gitignore/77e29837cf03b59fc4d885ea011bbd683caaaf85/node.gitignore
 
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 *.js
 *.js.map
 src/*.d.ts
-lib/
+dist/
 
 ### https://raw.github.com/github/gitignore/77e29837cf03b59fc4d885ea011bbd683caaaf85/node.gitignore
 
--- a/README.md
+++ b/README.md
@@ -29,8 +29,8 @@ Or Settings/Preferences ➔ Packages ➔ Search for `atom-beautify`
 
 ## Contribute
 1. Clone the repository
-1. `cd 'atom-beautify'`
-1. Run `npm install`
-1. Run `apm link` (if necessary)
-1. Install [atom-typescript](https://atom.io/packages/atom-typescript) (if using Atom)
-1. Typescript files in `/src` will be compiled to Javascript and saved in `/lib` whenever a file is saved
+2. `cd 'atom-beautify'`
+3. Run `npm install`
+4. Run `apm link` (if necessary)
+5. Install [atom-typescript](https://atom.io/packages/atom-typescript) (if using Atom)
+6. Typescript files in `/src` will be compiled to Javascript and saved in `/lib` whenever a file is saved
--- a/README.md
+++ b/README.md
@@ -26,3 +26,11 @@ apm install atom-beautify
 ```
 
 Or Settings/Preferences ➔ Packages ➔ Search for `atom-beautify`
+
+## Contribute
+1. Clone the repository
+1. `cd 'atom-beautify'`
+1. Run `npm install`
+1. Run `apm link` (if necessary)
+1. Install [atom-typescript](https://atom.io/packages/atom-typescript) (if using Atom)
+1. Typescript files in `/src` will be compiled to Javascript and saved in `/lib` whenever a file is saved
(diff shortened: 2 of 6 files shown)


szeck87:
Should probably comment out Travis builds for now.

Also, I felt it would be cleaner and faster to separate the .ts and .js files, that's why I have .js going to lib.  If that's not desirable and .js files should stay in src, I'll remove that line.

Glavin001:
> Also, I felt it would be cleaner and faster to separate the .ts and .js files, that's why I have .js going to lib. If that's not desirable and .js files should stay in src, I'll remove that line.

Agreed. Separating the TypeScript and generated JavaScript and type definition files (.d.ts) is ideal. I do not currently recall why they were setup like they were.. likely to make sure Travis CI passed 😝 .

Will need Travis CI passing before we can merge this.

.gitignore
```diff
@@ -2,6 +2,7 @@
 *.js
 *.js.map
 src/*.d.ts
+lib/
```
Glavin001: I usually prefer `dist/` since this is not really a library.

Glavin001: I usually prefer `dist/` since this is not really a library.

README.md
```diff
@@ -26,3 +26,11 @@ apm install atom-beautify
...
 
 Or Settings/Preferences ➔ Packages ➔ Search for `atom-beautify`
+
+## Contribute
+1. Clone the repository
+1. `cd 'atom-beautify'`
+1. Run `npm install`
+1. Run `apm link` (if necessary)
+1. Install [atom-typescript](https://atom.io/packages/atom-typescript) (if using Atom)
+1. Typescript files in `/src` will be compiled to Javascript and saved in `/lib` whenever a file is saved
```
Glavin001: The steps are all `1.`.

tsconfig.json
```diff
@@ -7,13 +8,15 @@
         "sourceMap": true,
         "target": "ES6",
         "declaration": true,
-        "moduleResolution": "node"
+        "moduleResolution": "node",
+        "outDir": "lib"
```
Glavin001: I usually prefer `dist/` since this is not really a library.

tsconfig.json
```diff
@@ -7,13 +8,15 @@
...
     },
     "include": [
         "./declarations.d.ts",
         "./src/**/*.ts",
         "./script/**/*.ts",
-        "./test/**/*.ts"
+        "./test/**/*.ts",
+        "src"
```
Glavin001: "src" should not be necessary since we already have `"./src/**/*.ts"`. Use only 1 of them.

.gitignore
szeck87: I updated it to use `dist/`

README.md
szeck87: The numbers for numbered lists don't matter in the raw file, it outputs as 1-6:  https://github.com/szeck87/atom-beautify/tree/stevenzunibeautify#contribute

tsconfig.json
szeck87: Updated to dist

tsconfig.json
szeck87: I was having trouble with the auto-compile and thought that was the problem, turns out it wasn't.  Removed it.

README.md
Glavin001: Yes, but what about uses -- like myself -- who use text editors such as [Vim](http://www.vim.org/) to view these README.md files without Markdown previews 😝 . The textual file should mean something, and then the Markdown rendering is a luxurious way to view it.

szeck87:
Updated readme

szeck87:
This ultimately compiles and Atom recognizes it as a package.  The weird thing that's happening is it's showing settings that don't exist anywhere in the project.

screen shot 2017-10-10 at 11 35 22 am

szeck87:
Some (all saved ones?) preferences for Atom Beautify are saved in config.cson (apart of the atom config), which is where these were coming from.  Something to keep in mind of if things shift there.

README.md
szeck87: Yea, but if you have to add a step in the middle, it's a pain to update.  I updated it here though.

Glavin001: > it's a pain to update.

Good. It will encourage us to keep the list of steps short 😜 .

Thanks.

Glavin001:
Good point about the previously saved configuration. Something to consider before we release.

Glavin001:
Thanks for contributing!

Glavin001:
Published to v0.30.6

PR #1897, opened 2017-10-20
autopep8: add option to respect user config

nathdwek:
Apparently for autopep8 CLI options and user config file (`.config/pep8`) are mutually exclusive. This means that the current behaviour of the package is that the package defaults will override the user defaults.

When `cli_options` is set to false, no style options are passed to the CLI so that autopep8 will respect the user config. Even though for this would make sense as the default behaviour, I left it to true by default so that backward compat is preserved.
Open questions:
* For now I went KISS, but it would be possible for the user to provide a custom pep8 config file via `--global-config`
* I don't know how to specify to which beautifier the option applies (only autopep8), so for now the option is labeled as applying to autopep8 and pybeautifier (but not yapf, I don't know why). This should be fixed before merging.


nathdwek:
Just in case this PR is being ignored because of CI issues: they are due to a totally unrelated rubygems issue.

It would be really nice if this simple change made its way into the plugin!

Issue #1869, opened 2017-09-29
Add Settings shortcut to the Packages menu


rishacha:
# Description
It would be really nice if you can add a shortcut to open the settings view as part of the package sub-menu.  
> This is my go-to tool before I submit any piece of code for review. Appreciate the work done :+1:

szeck87:
This is a pretty simple update.  I have it completed, just waiting on Linuxbrew to fix one of their formulas so Travis passes.  https://github.com/szeck87/atom-beautify/tree/add-settings-menu

Glavin001:
Published to v0.30.6

PR #1899, opened 2017-10-20
Add settings option to menu and use Ruby 2.1.10 in Travis

szeck87:
This adds an option to the Packages -> Atom Beautify submenu, Settings, that goes directly to the settings page for Atom Beautify.  This is also available in the command palette.

During the initial Travis build, Rubocop could not be installed on the OSX build because it requires Ruby 2.1 or greater.  The default is 2.0.0, thus a second change was added to use Ruby 2.1.10 so that the Travis build would pass (although it didn't fully, more on that below).
...

#1869 
...

### Any other comments?
The OSX Travis build will still fail.  I'm getting the same errors with puppet-lint and sass as the master branch here.  I'm still working with Travis support to try and resolve them, but the builds will continue to fail for now.  Since it doesn't fail until the very end, and the Linux builds are fine, it should be safe to merge IMO.
...

### Checklist

- [X] Add change details to `CHANGELOG.md` under "Next" section
- [X] AppVeyor passes (Windows support)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,6 @@ matrix:
     - os: osx
       env:
         - ATOM_CHANNEL=stable
-      rvm: 2.1.10
     # - os: osx
     #   env:
     #     - ATOM_CHANNEL=beta
@@ -98,7 +97,7 @@ before_install:
   - brew tap homebrew/versions
   # Ruby language support
   # - gem install ruby-beautify --verbose
-  - gem install rubocop
+  # - gem install rubocop
   - gem install htmlbeautifier
   - gem install puppet-lint
   # Sass language support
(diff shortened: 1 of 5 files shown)


szeck87:
@Glavin001 it looks like the Ruby unit tests are skipped, does Rubocop need to be installed in the Travis build?  If not, I'd rather comment it out and leave the Ruby version as is.

Glavin001:
@szeck87 , agreed. Comment out the Ruby change and let's merge.

Thanks for contributing!

Glavin001:
Thanks for contributing!

Glavin001:
Published to v0.30.6

PR #1916, opened 2017-10-30
Fix rustfmt-nightly version check.

kgv:
Nightly rustfmt version can contain 2 digits in PATCH field.
Current rustfmt-nightly version is: `0.2.13-nightly ( )`. And it throw the exception: "rustfmt version 0.5.0 or newer required".


alexheretic:
@kgv This looks good, I'm also having this issue.

Why not switch out the `?` for a `*` to handle more future versions, ie `0.2.164-nightly ( )`? 
```js
// match non-nightly pre 0.5 versions
/^0\.[0-4]\.[0-9](?![0-9]*-nightly)/
```

Issue #1951, opened 2017-11-22
Add option to use tabs instead of spaces for bash


Lawstorant:
beautysh supports tab indentation with an optional -t flag. It would be great if we had an option to use it with atom-beautify.

Lawstorant:
Figured it out and made a pull request.

Glavin001:
Published to v0.32.0

PR #1959, opened 2017-11-24
Added the option to indent Bash scripts with tabs

Lawstorant:
This change adds the ability to indent shell scripts with tabs. It does so by running beautysh with "-t" flag.

Closes https://github.com/Glavin001/atom-beautify/issues/1951

### DIFF
--- a/src/languages/bash.coffee
+++ b/src/languages/bash.coffee
@@ -27,5 +27,9 @@ module.exports = {
       default: null
       minimum: 0
       description: "Indentation size/length"
+    use_tabs:
+      type: 'boolean'
+      default: null
+      description: "Use tabs for indentation"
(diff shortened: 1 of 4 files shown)


src/languages/bash.coffee
```diff
@@ -27,5 +27,9 @@ module.exports = {
       default: null
       minimum: 0
       description: "Indentation size/length"
+    use_tabs:
```
Glavin001: We try to keep the option names the same as other languages. In this case, `indent_with_tabs` would be the appropriate option name which we already use.

Glavin001:
Thank you for contributing!

Glavin001:
Published to v0.32.0

PR #1999, opened 2018-01-05
Cleanup build process

stephengroat:

Uses `Gemfile` for ruby dependencies

Implmements the use of a `Brewfile` and [`brew bundle`](https://github.com/Homebrew/homebrew-bundle), moving `osx` dependencies out of the travis YAML file and into a `Brewfile`

also uses a python `requirements.txt` to move pip dependencies into a single location

simplify `travis` file, `trusty` is now the default, no need to explicitly state it

...

...

### Any other comments?

...

### Checklist


szeck87:
Hi @stephengroat, can you merge your branch with master?  I just merged a fix for the AppVeyor and Travis OSX build issues (you'll see pip2 instead of pip for OSX).  After that push your changes with it and let the build run again.

szeck87:
Wondering if linuxbrew even needs to be installed.  Would probably have to fix the paths for php-cs-fixer and elm-format, but that should be pretty easy.

stephengroat:
@szeck87 instead of `linuxbrew`, which hasn't been that popular, you could use `nix`, which seems to be getting a lot more support (for `linux` at least). You could probably get rid of most of the docker images that way (if your interested). [here's an example of a repo using nix pretty well](https://github.com/UltimateHackingKeyboard/firmware/blob/master/.travis.yml#L20-L22)

i can try to do that, but, if possible, i'd like to make that a seperate pr

stephengroat:
@szeck87 i did a sample `nix` implementation, let me know what you think

szeck87:
@stephengroat I like it.  Simplifies the appveyor and travis config files.  I went ahead and cancelled builds not from your latest commit so Travis doesn't get backed up (it happens a lot with their OSX capacity).  Just a few things:

* Please update any comments in the Travis config file that need it.  e.g. https://github.com/stephengroat/atom-beautify/blob/patch-1/.travis.yml#L60 would need to reflect that it's installing nix and not Homebrew on Linux.
* We probably should test out the Docker images to ensure the beautifiers there work properly, but I think it's better to do that separately (different job stages?).  @Glavin001 I don't know what your thoughts are on that.
* I haven't worked very much with separate files for these package managers, but when users install atom-beautify it won't install the packages in the Brewfile, Gemfile, etc correct?  Want to ensure that at least for now they're only used for the CI builds.

stephengroat:
@szeck87 thanks. could you also kill the currently running `appveyor` build? i forgot a `-y`, it'll have to timeout otherwise

szeck87:
@stephengroat I tried and I can't, the access there is a bit different than Travis.

stephengroat:
@szeck87 no problem, thanks for trying!

szeck87:
@stephengroat also the build-package.sh script installs Atom, so if you're installing it through a cask then it will be duplicated (or there will be two of them when running the Beta build since there is no formula for the Beta).  Secondly, the `then brew update && brew cask zap oclint && brew bundle && ln -s /usr/local/bin/elm-format-0.17 /usr/local/bin/elm-format` script takes 11 minutes to run.  I wonder if we can cache the Homebrew dependencies, can you try adding `$HOME/Library/Caches/Homebrew` under cache -> directories?

stephengroat:
@szeck87 added.

The worst part of that command is `brew update`, takes ~5 mins
`brew bundle` takes another 4-5 mins on install
I've tried that trick before, it does same time if the downloads are large, but the disk speed on `osx` is pretty bad, and with `brew`, there's a ton of reads/writes with `git`, etc, so it's always a bit slower

szeck87:
@stephengroat yea I know.  `brew update` (as of right now) takes around 2 minutes since the image is somewhat fresh.  I think it's spending the majority of the time in `brew bundle` (maybe 7-8 minutes).  I don't know if the caching will help (or if it will cache at all), but worth a try.  I think that's also why the beta build for OSX was not put in, so it would speed up the build time.

szeck87:
@stephengroat looks like that Homebrew line causes caching storing to timeout, go ahead and remove it.  Sorry about that.

stephengroat:
need to bump to new version of uncrustify https://github.com/StevenWolfe/choco-uncrustify/pull/1

szeck87:
@stephengroat I can cancel appveyor builds now, just tag me comment if you need it.

stephengroat:
@szeck87 gotcha.

do you want `atom` and `atom-beta` installed via `brew` and `nix`? i saw that `appveyor` uses `chocolately` for it, but i'm not sure if you want to use the package managers for the other methods as well

szeck87:
@stephengroat I think right now both Atom and Atom Beta are installed via the build-package.sh before the vars are set.  I'm all for cleaner code which putting them in package managers should do, as long as it's not at the expense of performance and the tests run.

stephengroat:
@szeck87 the `linux` and `osx` paths on `travis` are solid and ready to go

the `appveyor` path is solid, just missing one binary, i'm working on getting it updated https://github.com/StevenWolfe/choco-uncrustify/pull/1

comment away, let me know if you have any questions

could you tell me what i lost on "Better Code Hub"? I can't see which guideline i lost

szeck87:
@stephengroat looking good.  I wonder if the Homebrew dependencies are automatically cached, since the latest build took roughly half the time compared to your earlier builds.  Or the CPUs for it were just taxed out.

I'm not sure how much luck you'll have with getting choco-uncrustify updated, it looks like it hasn't been maintained (the first and last commits were in December 2016), but worth a try I guess.

I can't tell what the issue with Better Code Hub is, @Glavin001 would have to look at it.

Glavin001:
Let's ignore the "Better Code Hub" failure for now. This is an awesome Pull Request and little improvements can be pushed to later Pull Requests, if needed. 😃 

Thank you @stephengroat and @szeck87 both for your work on this! 🎆 🎉

szeck87:
It's all @stephengroat.  Let us know when you're complete with commits and we'll review it.  I'll look into Better Code Hub in the meantime.

szeck87:
Once this is merged, the Better Code Hub guidelines here should be the new "default" for future PRs (hopefully).  That way it won't fail for every new PR.

szeck87:
@stephengroat installing ghc and msys2 makes the appveyor build take an extra 15 minutes to run and the page freezes when I try and view the details (guessing too many lines on the page).  Does stylish-haskell not install with just haskell-stack?  I think we're skipping the Haskell unit tests right now, but it would be good to have it installed so we can start to not skip them.

.travis.yml
```diff
@@ -85,110 +54,41 @@ cache:
...
-      pip install --user --upgrade sqlparse;
-    else
-      pip2 install --upgrade sqlparse;
-    fi
-  # Java, C, C++, C#, Objective-C, D, Pawn, Vala
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      brew install uncrustify;
-    else
-      docker pull unibeautify/uncrustify;
+      echo Error:TRAVIS_OS_NAME && exit 1;
```
Glavin001: I think this should be:

```bash
echo "Error: $TRAVIS_OS_NAME" &&  && exit 1;
```

.travis.yml
```diff
@@ -85,110 +54,41 @@ cache:
...
-  - gem install puppet-lint
-  # Sass language support
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      gem install sass;
-    else
-      docker pull unibeautify/sass-convert;
-    fi
-  # Python language support
+    # Linux: Install Nix and Packages, Python install
+    # OSX: Update Homebrew, remove conflicting cask, brew bundle, symlink elm-format
```
Glavin001: How about moving these comments closer to where these things happen?
Or remove them entirely. I don't want to see them become out of date in future commits 😉 .

.travis.yml
stephengroat: why the double `&&`? i can't get that to work

.travis.yml
stephengroat: super hard to do with the `if` statements, `travis` has a really weird/slightly undocumented way of interpreting/escape `bash` in it's `yaml`. see https://travis-ci.org/Glavin001/atom-beautify/builds/327062135

do you want to separate that script out into a seperate file?

.travis.yml
szeck87: Yea don't need the double ampersands, but need `"Error: $TRAVIS_OS_NAME"` in quotes.

stephengroat: @szeck87 still invalid `yaml` https://travis-ci.org/Glavin001/atom-beautify/builds/327070475

stephengroat:
@Glavin001 `yaml` issue fixed

.travis.yml
szeck87: Yea that's odd, I thought you had to put quotes when echoing a string.  Sorry about that.  That line probably won't ever be hit, at least in the near future, as the only OS names are linux and osx.

szeck87:
@Glavin001 @stephengroat just my comment from before where the AppVeyor build basically crashes my browser when I try and view it.  I believe it has something to do with the ghc and/or msys2 installs.  We'll need to be able to view those builds if they fail in the future to determine what problem is.  Is it just for Haskell?  Is haskell-stack not enough?  If so, can we just skip the Haskell/ghc/msys2 installs for Windows for the time being?

stephengroat:
@szeck87 exactly, it's the `haskell-stack` install. `haskell-stack` decompresses a **TON** of stuff during setup and 7zip doesn't have a non-verbose option, so it floods to terminal.

do you want to only do haskell on `linux`?

szeck87:
@stephengroat can you comment the installs for Haskell out for all three?  It's fine to leave the code there for future purposes, but those tests are being skipped right now anyways (hence the underscore in the original/expected test files for Haskell), and the install issues is probably why.

stephengroat:
cleaned up and ready to merge

szeck87:
I cleared the Travis cache as this PR had almost a gig cached throughout it's lifecycle (and probably caused that ocaml error).  Waiting for the build to pass then I'll approve my review.

szeck87:
@Glavin001 can this be merged?

.travis.yml
Glavin001: Oops. Typo on my end. Thanks for correcting!

Glavin001:
Absolutely awesome work @stephengroat!! Thank you!

I think this is ready to merge 🎉 .

stephengroat:
@Glavin001 @szeck87 no problem. thanks for all the great work you guys continue to do!

szeck87:
The OSX builds are still failing, one times out because nothing is written to the console for 10 minutes and thus times out.  The other due to an error with sass and puppet-lint, both installed from bundler.  I deleted the cache and restarted it yesterday and ended up with the same result.  A few observations:

1. Travis' OSX builds are pretty messed up right now, as they're running at half the usual capacity and some of the services are unstable.

2. Since there is a Gemfile, Travis also runs `bundle install --jobs=3 --retry=3 --path=${BUNDLE_PATH:-vendor/bundle}`, which is also run earlier in the before_install step.  But Linux builds are fine.

3. A PR was opened after this was merged, and that PR built successfully, so it could just be an issue with master branch.

stephengroat:
@szeck87 if you want to switch the build language to `generic`, that'll stop the re-run of the `Gemfile`

szeck87:
I updated my fork to comment out the `bundle install` command as well as the cache directory for it.  According to Travis' [docs](https://docs.travis-ci.com/user/caching/#Determining-the-bundle-path), it should automatically determine where to put/pull cache.  But still don't even know if that's the issue, my build has been going for 3 hours now and OSX still hasn't started.

szeck87:
OK, I just changed it to `language: generic` and that worked.  So the non-cached builds took:

16:12
16:41
18:23
17:41

Let's see what happens tomorrow during the next run from cached resources.

Issue #965, opened 2016-05-04
.remarkrc file is not respected


chdsbd:
### Description

When beautifying markdown using remark, atom-beautify does not respect .remarkrc file in the home directory.
### Expected Results

Atom-beautify should follow the .remarkrc file in home directory by default, just as linter-markdown does.
### Steps to Reproduce
1. Add code to Atom editor
2. Run command `Atom Beautify: Beautify Editor`
3. This beautified code does not look right!

ChrisRast:
Hi @Glavin001, what can I do to address this issue? I have the same problem...
Can you give us indication on where to look if we want to add support for a configuration file for remark?
Thank you for your work!

Glavin001:
Hi @ChrisRast.

Pull Requests welcome! Applicable file is https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/remark.coffee#L39

Please feel free to open the PR earlier than later so I can review and answer questions 😃 .

ChrisRast:
Still an issue, but did not have time & knowledge to address it. :(

kachkaev:
@Glavin001 I'm interested in implementing this feature and submitting a PR. However, there is an issue I'd like to ask @wooorm, the author of remark. It is to do with how `.remarkrc` is searched.

It's great that [`beautify` method](https://github.com/Glavin001/atom-beautify/blob/master/src/beautifiers/remark.coffee#L35) can receive the fourth parameter called `context`, which means that we have access to `context.filePath` and therefore `path.dirname(context.filePath)`. This is where we should start searching for `.remarkrc` or `.remarkrc.js`. Glancing through `remark` and its parent `unifiedjs` lib I noticed the use of `cwd`, which suggests that this is how _current working directory_ could be defined. However, none of my experiments to set it worked. These are the most sane attempts:

```coffee
modifiedOptions = Object.assign({}, options, {detectConfig: true })
cleanMarkdown = remark({
  cwd: path.dirname(context.filePath)
}).process(text, modifiedOptions).toString()
```

```coffee
modifiedOptions = Object.assign({}, options, {detectConfig: true, cwd: path.dirname(context.filePath) })
cleanMarkdown = remark().process(text, modifiedOptions).toString()
```
@wooorm could you please hint on how to pass `cwd` into remark so that it could find `remarkrc`?

kachkaev:
Ping @wooorm. Will appreciate your help here 🙌

wooorm:
[...]

PR #744, opened 2016-01-13
PHPCBF (PHP_CodeSniffer) + Remark for Markdown support

ra100:
Quick support for phpcbf related to issue https://github.com/Glavin001/atom-beautify/issues/343

PR #2004, opened 2018-01-09
Use Remark beautifier for markdown files by default

kachkaev:

Tidy Markdown was the first and only markdown beautifier in `atom-beautify` until [Remark](https://github.com/remarkjs/remark) was [added](https://github.com/Glavin001/atom-beautify/pull/744) two years ago. As of January 2018, Remark seems to be a more promising solution for markdown according to the state of its community:

| | Tidy Markdown | Remark |
|-|---|---|
|number of contributors | [2](https://github.com/slang800/tidy-markdown/graphs/contributors) | [60](https://github.com/remarkjs/remark/graphs/contributors)
commits to master in 2017 | [11](https://github.com/slang800/tidy-markdown/commits/master) | [50+](https://github.com/remarkjs/remark/commits/master)
|github stars| [35](https://github.com/slang800/tidy-markdown/stargazers) | [1235](https://github.com/remarkjs/remark)
| downloads per months on npm | [45K+](https://www.npmjs.com/package/tidy-markdown) | [660K+](https://www.npmjs.com/package/remark)

I'd like to propose switching to Remark by default and thus serve `atom-beautify` users with something they are more likely to be familiar with.

This may positively affect https://github.com/Glavin001/atom-beautify/pull/1990 because Tidy Markdown has some issues extracting code from code blocks. If Remark is made a default choice, we can simply disable code block beautification option for Tidy Markdown without any harm to most users.

Also, if the solution to https://github.com/Glavin001/atom-beautify/issues/965 is found, the default markdown beautifier will better integrate with custom markdown preferences that users may have.

### Checklist


kachkaev:
Tests also need updating because of slightly different default behaviours of Remark and Tidy Markdown

szeck87:
I'm not against this, but our remark version is really out of date (3 major versions behind).  Between this, #2008 and #1990 it might be better to make it one PR since these files are all crossing one another.  Also, it's preferable to use node.js for beautifiers if available so users don't have to install them separately. I would suggest looking for a global install of remark-cli first, then if there isn't one, fall back to the programmatic interface of Remark (which needs to be upgraded in atom-beautify).

kachkaev:
@szeck87 so which PR would you recommend to leave as an umbrella for all these changes? I've been thinking that they deserve separate discussions and that only some suggestions will be accepted. I'll be happy to restructure my work if you can recommend what to do.

kachkaev:
Checking for `remark-cli` and falling back to a built-in remark via npm seems reasonable. Do we need any ticks in options or should I just search for `node_modules/.bin/remark-cli`, then `remark-cli` and finally use `require('remark-cli')` on failure?

szeck87:
@Glavin001 what's your opinion on the PR structures for #2004, #2008, and #1990?

As far as checking for global versus options, I seem to remember working on something similar and tried to use [requireg](https://www.npmjs.com/package/requireg), but can't remember what became of it.  Is the use of remark-cli just for use of a .remarkrc file?  @wooorm does node version not support the use of those files?

wooorm:
@szeck87 Sorry, I’m not entirely sure what your question is?

szeck87:
@wooorm yea sorry, let me start over.  In your comment on #965, you suggested to use remark-cli instead of remark.  What is the rationale there?

wooorm:
* `remark-cli` (based on unified-engine) supports file finding, config files, file ignoring, etc etc: it’s made for the file-system, and has a higher level interface.
* `remark` (based on unified) is made for processing text, and to be used through code

szeck87:
On second thought @kachkaev I think these can all stay as separate PRs.  But I would hold off on #1990, or at least the remark specific parts of it, until #2008 is finalized.

szeck87:
@kachkaev this PR needs a couple more things:

* Can you update the link in remark.coffee to https://github.com/remarkjs/remark?
* Update the examples under `examples/nested-jsbeautifyrc/markdown/expected/test.md` and `examples/simple-jsbeautifyrc/markdown/expected/test.md` and `yaml-front-matter.md`?  The other two yaml named files are skipped so you can ignore them.

kachkaev:
@szeck87 done 👍

examples/simple-jsbeautifyrc/markdown/expected/yaml-front-matter.md
```diff
@@ -1,7 +1,8 @@
...
-title: this is my title
-description: this is my description
+layout:          default
+title:        this is my title
+description:     this is my description
 ---
 
-test ing
+ test
+   ing
```
Glavin001: It looks like Remark simply did NOT beautify this file.
Original: https://raw.githubusercontent.com/kachkaev/atom-beautify/18e7f5ebad5f83e25a6a03e05ee15ada93329328/examples/simple-jsbeautifyrc/markdown/original/yaml-front-matter.md

```markdown
---
layout:          default
title:        this is my title
description:     this is my description
---

 test
   ing
```

examples/simple-jsbeautifyrc/markdown/expected/test.md
```diff
@@ -1,11 +1,11 @@
...
+-   item
 
 ## heading 2
 
-1. one
-2. two
-3. three
+1.  one
+2.  two
+3.  three
```
Glavin001: Looks like it worked. However, how can a user configure this indentation? It appears to be 3 spaces for `-` bullets and 2 spaces for numbered bullets, such that they align for both. This looks nice, however for our current users this will cause a lot of whitespace reformatting when they beautify their Markdown files. If we could configure the default to work like it did previous that would be ideal.

Original: https://raw.githubusercontent.com/kachkaev/atom-beautify/18e7f5ebad5f83e25a6a03e05ee15ada93329328/examples/simple-jsbeautifyrc/markdown/original/test.md

```markdown
#            heading 1

- item
-  item
-   item

##     heading 2

1. one
2.  two
2.   three
```

src/languages/markdown.coffee
```diff
@@ -18,7 +18,7 @@ module.exports = {
     "md"
   ]
 
-  defaultBeautifier: "Tidy Markdown"
+  defaultBeautifier: "Remark"
```
Glavin001: I think you make a good argument for defaulting to Remark 👍 . However, first need to address the changes in `expected` above.

Glavin001:
Thank you for this Pull Request, @kachkaev! Great work so far. Close to merging 😄 .

src/beautifiers/remark.coffee
```diff
@@ -3,7 +3,7 @@ Beautifier = require('./beautifier')
 
 module.exports = class Remark extends Beautifier
   name: "Remark"
-  link: "https://github.com/wooorm/remark"
+  link: "https://github.com/remarkjs/remark"
```
Glavin001: Options are listed: https://github.com/remarkjs/remark/tree/master/packages/remark-stringify#options

examples/simple-jsbeautifyrc/markdown/expected/test.md
Glavin001: I think this option is `listItemIndent` or `incrementListMarker`: https://github.com/remarkjs/remark/tree/master/packages/remark-stringify#optionslistitemindent

examples/simple-jsbeautifyrc/markdown/expected/yaml-front-matter.md
kachkaev: Yes, unfortunately remark does not format `frontmatter` and also does not collate the lines like in the example above. Shall we just remove this test for now?

examples/simple-jsbeautifyrc/markdown/expected/test.md
kachkaev: Hmm... I'm not sure we should be switching back to old indents. Yes, the existing users may experience a one-time problem after an update, but that's expected given that the formatter being used is now different. Anyone can switch back to _Tidy Markdown_ and thus restore the original behaviour.

Four characters per intent is a very reasonable choice and sticking to this rule has more benefits in the long run. In any case, it's not the only breaking change that people might see.

examples/simple-jsbeautifyrc/markdown/expected/yaml-front-matter.md
Glavin001: No, this would become a regression bug for Atom-Beautify. If our users assumed continued support for YAML front-matter formatting in their Markdown files, then switching the default will completely break their usage.

I'm leaning away from switching default to Remark unless we can beautify files as least as well as we had done previous with the default (TidyMarkdown). Next release should be a clear improvement, not a bug, to users.

kachkaev: @wooorm do you think remark could solve this somehow?

examples/simple-jsbeautifyrc/markdown/expected/test.md
Glavin001: In light of the lack of YAML font-matter support and these indentation changes, we need to make sure the `CHANGELOG` describes all of the breaking changes made by switching to Remark. Also, I would like to have *more* examples/tests added for Markdown showcasing the value and expected result from switching to Remark.

examples/simple-jsbeautifyrc/markdown/expected/yaml-front-matter.md
Glavin001: @kachkaev could you describe the possible breaking changes in the CHANGELOG? If we are going to switch from Tidy Markdown to Remark and cause all of these issues, we need to make it clear the cause and solution for users who do like their formatting looking like before.

kachkaev:
Thanks for your feedback @Glavin001! I tried to address the questions you've raised and am happy to hear what you think again!

kachkaev:
CI fails, but according to the logs, the problem is with `examples/simple-jsbeautifyrc/tsx/expected/test.tsx`, not markdown files 🤔

examples/simple-jsbeautifyrc/markdown/expected/yaml-front-matter.md
wooorm: What is “frontmatter”? Here you show YAML, but there also exists TOML and some JSON front matter format as well.

Yes, remark could support this, through plugins, something like https://github.com/remarkjs/ideas/issues/3, and the linked issue contains some sample code, that should be made to format the YAML/TOML/JSON!

kachkaev:
Hmm... After checking out markdown formatting with [prettier](https://github.com/prettier/prettier) in the last couple of days, I'm in doubt that we should switch from 2/3 space indents to a fixed indent of four. Prettier also uses remark, but it's only for parsing. The details on how they generate the output is in this feature PR:
https://github.com/prettier/prettier/pull/2943

So what shall we do with list item spacing? 🤔

Glavin001:
Thank you for contributing and your great work, @kachkaev !!

Glavin001:
Published to v0.32.0

PR #2008, opened 2018-01-11
Switch to remark-cli to support .remarkrc and unlock remark version

### LINKED ISSUE #965: .remarkrc file is not respected

kachkaev:

I've been trying to add `.remarkrc` support to fix #965 and was [suggested](https://github.com/Glavin001/atom-beautify/issues/965#issuecomment-356340663) by @wooorm to have a look at `remark-cli`. As a result, I re-implemented remark beautifier to use `remark` as an external tool.

The benefit of this approach is not only a local `.remarkrc` file can be found and loaded, but also that there is no more version lock-in for `remark` (it used to be v6, which is 2 versions behind the latest).

I don't pretend that this PR is perfect and will be happy to revise it with someone's help.

Closes https://github.com/Glavin001/atom-beautify/issues/965. Related to https://github.com/Glavin001/atom-beautify/pull/2004.

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -185,7 +185,6 @@
     "open": "0.0.5",
     "prettydiff2": "^2.2.7",
     "pug-beautify": "^0.1.1",
-    "remark": "6.0.1",
     "season": "6.0.0",
     "semver": "^5.3.0",
     "shell-env": "^0.3.0",
--- a/src/beautifiers/remark.coffee
+++ b/src/beautifiers/remark.coffee
@@ -1,44 +1,62 @@
+###
+Requires https://github.com/remarkjs/remark
+###
 "use strict"
 Beautifier = require('./beautifier')
 
 module.exports = class Remark extends Beautifier
   name: "Remark"
-  link: "https://github.com/wooorm/remark"
+  link: "https://github.com/remarkjs/remark"
   options: {
     _: {
       gfm: true
       yaml: true
       commonmark: true
-      footnotes: true
-      pedantic: true
-      breaks: true
-      entities: true
-      setext: true
-      closeAtx: true
-      looseTable: true
-      spacedTable: true
-      fence: true
-      fences: true
-      bullet: true
-      listItemIndent: true
-      incrementListMarker: true
-      rule: true
-      ruleRepetition: true
-      ruleSpaces: true
-      strong: true
-      emphasis: true
-      position: true
+      search_for_configuration: true
     }
     Markdown: true
   }
 
-  beautify: (text, language, options) ->
-    return new @Promise((resolve, reject) ->
-      try
-        remark = require 'remark'
-        cleanMarkdown = remark().process(text, options).toString()
-        resolve cleanMarkdown
-      catch err
-        @error("Remark error: #{err}")
-        reject(err)
+  executables: [
+    {
+      name: "Remark CLI"
+      cmd: "remark"
+      homepage: "https://github.com/remarkjs/remark"
+      installation: "https://github.com/remarkjs/remark/tree/master/packages/remark-cli#installation"
+      version: {
+        parse: (text) ->
+          return text.match(/remark: (\d+\.\d+\.\d+)/)[1]
+        runOptions: {
+          returnStderr: true
+        }
+      }
+    }
+  ]
+
+  beautify: (text, language, options, context) ->
+    cwd = context.filePath and path.dirname context.filePath
+    remarkArgs = [
+      '--no-color'
+      '--no-ignore'
+      '--file-path'
+      context.filePath
+    ]
+    remarkOptions = {}
+
+    if (!options.search_for_configuration)
+      settings = Object.assign({}, options)
+      delete settings.search_for_configuration
+      serializedSettings = JSON.stringify(settings).slice(1, -1)
+      remarkArgs.push('--no-config', '--setting', serializedSettings)
+
+    return new Promise((resolve, reject) =>
+      @exe("remark")
+        .run(remarkArgs, {
+          cwd
+          onStdin: (stdin) ->
+            stdin.end text
+        })
+        .then (output) ->
+          resolve(output)
+        .catch(reject)
     )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Next
+- Switch to external [`Remark CLI`](https://github.com/remarkjs/remark/tree/master/packages/remark-cli) and thus allow for user-defined config with `remarkrc` ([#965](https://github.com/Glavin001/atom-beautify/issues/965))
 - Fix [#1862](https://github.com/Glavin001/atom-beautify/issues/1862) Add support for ocp-indent as an executable
 
 # v0.30.9 (2017-11-22)
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ Some of the supported beautifiers are developed for Node.js and are automaticall
 | Pug Beautify | :white_check_mark: | :ok_hand: Not necessary | :smiley: Nothing! |
 | puppet-lint | :warning: 1 executable | :white_check_mark: :100:% of executables | :whale: With [Docker](https://www.docker.com/):<br/>1. Install [puppet-lint (`puppet-lint`)](http://puppet-lint.com/) with `docker pull unibeautify/puppet-lint`<br/><br/>:bookmark_tabs: Manually:<br/>1. Install [puppet-lint (`puppet-lint`)](http://puppet-lint.com/) by following http://puppet-lint.com/<br/> |
 | pybeautifier | :warning: Manual installation | :construction: Not an executable | :page_facing_up: Go to https://github.com/guyskk/pybeautifier and follow the instructions. |
-| Remark | :white_check_mark: | :ok_hand: Not necessary | :smiley: Nothing! |
+| Remark | :warning: 1 executable | :x: No Docker support | :bookmark_tabs: Manually:<br/>1. Install [Remark CLI (`remark`)](https://github.com/remarkjs/remark) by following https://github.com/remarkjs/remark/tree/master/packages/remark-cli#installation<br/> |
 | Rubocop | :warning: Manual installation | :construction: Not an executable | :page_facing_up: Go to https://github.com/bbatsov/rubocop and follow the instructions. |
 | Ruby Beautify | :warning: Manual installation | :construction: Not an executable | :page_facing_up: Go to https://github.com/erniebrodeur/ruby-beautify and follow the instructions. |
 | rustfmt | :warning: Manual installation | :construction: Not an executable | :page_facing_up: Go to https://github.com/rust-lang-nursery/rustfmt and follow the instructions. |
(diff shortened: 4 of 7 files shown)


kachkaev:
@wooorm there's one thing that's working not as I've been expecting and I'm not sure if this is a peculiarity of `remark` or if I'm just missing anything in `remark-cli` args.

When I define a rule (e.g. `[require('remark-lint-fenced-code-marker'), '~']` inside my `.remarkrc.js`, I start seeing warnings related to this rule in [atom's interface](https://atom.io/packages/linter-markdown) and the command line, which is great. In this case I also expect `remark` to fix the rule if possible, just as this happens for `eslint --fix`. However, although the rule is detected by `remark-cli` and the warning is shown, certain fixable characters remain unchanged and the warnings persists. I've seen that `remark` has got a stringify module that takes part in forming the output, but I'm not sure how to make it simply follow all the existing linting rules without any extra configuration.

What would you suggest?

src/beautifiers/remark.coffee
```diff
@@ -1,44 +1,62 @@
+###
+Requires https://github.com/remarkjs/remark
+###
 "use strict"
 Beautifier = require('./beautifier')
 
 module.exports = class Remark extends Beautifier
   name: "Remark"
-  link: "https://github.com/wooorm/remark"
+  link: "https://github.com/remarkjs/remark"
```
kachkaev: The project has been moved to an org in ≈ Dec 2017

src/beautifiers/remark.coffee
```diff
@@ -1,44 +1,62 @@
...
+    ]
+    remarkOptions = {}
+
+    if (!options.search_for_configuration)
+      settings = Object.assign({}, options)
+      delete settings.search_for_configuration
+      serializedSettings = JSON.stringify(settings).slice(1, -1)
+      remarkArgs.push('--no-config', '--setting', serializedSettings)
+
+    return new Promise((resolve, reject) =>
```
kachkaev: A new promise object is not really necessary here at this point, but it is helpful to integrate the new implementation of the beautifier with #1990

wooorm:
> When I define a rule (e.g. [require('remark-lint-fenced-code-marker'), '~'] inside my .remarkrc.js, I start seeing warnings related to this rule in atom's interface and the command line, which is great. In this case I also expect remark to fix the rule if possible, just as this happens for eslint --fix. However, although the rule is detected by remark-cli and the warning is shown, certain fixable characters remain unchanged and the warnings persists. I've seen that remark has got a stringify module that takes part in forming the output, but I'm not sure how to make it simply follow all the existing linting rules without any extra configuration.

Yes, remark-lint just warns, it doesn’t change anything. remark-stringify (used in remark-cli, by the remark library itself, and here as well) does support some settings to beautify a certain way.
Some rules are aligned with remark-stringify settings. Others are not, and as of yet require a human to change something.

There’s also no code yet to map lint rule settings to remark-stringify settings, but that would be useful for other tools as well!

Alternatively, prettier can now format markdown as well, and it’s based on remark. It has a very nice line wrapping algorithm, that remark itself doesn’t have yet!

Does that help? Let me know if I can be of more help!

kachkaev:
Thanks for the explanation @wooorm. Hmmm I did not know that prettier can work with markdown via remark, interesting.

To be honest, my overall goal is to come up with an environment that’s similar to what prettier-eslint provides. There I’ve got some rules defined in `.eslint`, which influence the result of auto-formatting. What

kachkaev:
ping @wooorm :–)

wooorm:
Sorry for the slow response! You need some custom function, I guess? I’m not sure, it would need some work figuring out what remark-stringify settings correspond to remark-lint rules. How can I help?

szeck87:
@wooorm sorry I was tagging you in the wrong thread, it should have been here.  Can remark (based on unified) format markdown and support a `.remarkrc` file for configuration?  Or is the latter only available in remark-cli?

wooorm:
.remarkrc files are only supported by remark-cli. See [`remark-cli`](https://github.com/remarkjs/remark/tree/master/packages/remark-cli) for more details!

kachkaev:
@wooorm  so can we somehow use `.remarkrc` to guide `remark-cli`? Just wondering what to do with this PR. It was supposed to switch users to an external `remark-cli` and thus give them control over the result of formatting with `.remarkrc`, but now I'm not sure about its usefulness 😆

wooorm:
Well, the `settings` could be supported, and if they were, would allow humans to define how atom-beatify formats markdown through their `.remarkrc`

You could also allow the plugins. That could allow weird things though. Say someone has a pipeline to transform markdown to HTML, if atom-beautify respected that config, all files would end up as HTML!

But first: *why* do you want `.remarkrc` files?

kachkaev:
I'd like to enable the same functionality as [prettier-eslint](https://github.com/prettier/prettier-eslint) provides. That is to allow a team to deviate from 'industry standard' linting rules but still be able to autoformat the source files. Someone's editing policy may require `**` for bold text but for someone else a better choice is `__` for their own reasons.

When I contribute to some project docs, I'd like to be able to make a linter happy in one key press rather than by fiddling with micro-editing of symbols and spaces. That's when `atom-beautify` together with `.remarkrc` would be a  perfect tool.  Because `.remarkrc` exists in some repos for running `npm run lint:md` in CI already, it feels natural to employ the same file during beautification as well (just like `.eslintrc` is used by `npm run lint:js` and `prettier-eslint` inside Atom).

Of course, in an ideal world we would all be just using some unified formatting rules, but unfortunately we're quite far from that. Prettier has been trying to do standardise JS files for some time but that still does not work well and ends up with lots holy wars in [their issues on github](https://github.com/prettier/prettier/issues) 😄

wooorm:
Well, in that case `.remarkrc` is the solution right?

Now, the `settings` just supply info the the parser (e.g., footnotes support) and compiler (aka stringifier, like which marker to use for lists, `*`, `+`, or `-`).

`plugins` are also found in `.remarkrc` files, such as `remark-toc`, to autogenerate a table of contents if a `Table of Contents` header is found.

Other plugins, like the `remark-lint` rules, just warn about things.

People are free to set up remark lint rules exactly like the compilation settings (or not, but that’s weird). However, I don’t think that should be a concern to `atom-beautify`.

Anyway: the best way to do this is by depending on `unified-engine` I think!

kachkaev:
I agree that `.remarkrc` is a solution. I'm just struggling to figure out how to make `remark-cli` read the plugins and rules from `.remarkrc` and then `--output` the result, but without applying `toc` or other breaking changes like converting everything to HTML. It feels like this feature may naturally be a part of `remark`, and so atom-beautify will just calls this command with specific arguments. Actually, that's what already happening in this PR :–)

wooorm:
What’s wrong with `remark-toc`, if people have that set up? “breaking changes like converting everything to HTML”, who is to decide what constitutes a breaking change, and what an “okay” change like fixing some line-wrapping for example?
I feel that if people have “breaking things” in their `.remarkrc`, they explicitly did so 🤷‍♂️

kachkaev:
Could it be that they added `remark-toc` etc. to `.remarkrc` to export the entire project somewhere, but then want to simply fix their linting warnings? These seem like tasks with different expected outcomes, but with a shared config (at least partially). Fixing linting errors in markdown is similar to `eslint --fix` (i.e. get the list of errors and warnings and do whatever possible to reduce them).

szeck87:
@kachkaev I checked out your branch locally and, apart from `cwd` being invalid, it looks to read from a `.remarkrc` file as you intended.  What specifically are you looking to do at this point (or give an example of what your setup is and what it's doing that you don't)?

wooorm:
You could make a `.remarkrc` config file do anything, but `remark-cli` and `.remarkrc` files only really make sense if you’re working with markdown input *and* output.

I have both formatting stuff like remark-toc (`eslint --fix`) and linting stuff like remark lint (eslint’s unfixable rules). In my opinion, `remark-toc` updates my table of contents according to rules I have set up. I feel that that plugin is just as much like `eslint --fix` and `prettier`, in that it fixes my table of contents!

kachkaev:
> I feel that that plugin is just as much like eslint --fix and prettier, in that it fixes my table of contents!

@wooorm yeah the `toc` plugin is perhaps something similar to `eslint --fix` indeed. I did not think that TOC can re-generate itself multiple times without breaking things when a TOC has already been generated, but if that's how it works, great. However, we can't say the same about html-generating plugins, unfortunately. In any case, turning `.remarkrc` statements like `[require('remark-lint-heading-style'), 'setext']` or `[require('remark-lint-ordered-list-marker-value'), 'one']` into some kind of `eslint --fix` still feels like a missing feature.

@szeck87 `.remarkrc` is being read indeed, it's just not working as I originally expected because linting rules do not affect the beautification really 🤔

kachkaev:
A side-thought: is any preference given to a locally installed `remark-cli`? Or does `@exe("remark")` grab a globally installed instance in call cases?

szeck87:
@kachkaev It won't read any locally installed version at all.  I believe it is _possible_ to do so, but it requires a lot of new code since it checks the version of the executable before it even hits the beautify function.

kachkaev:
Right. Do other beautifiers give preference to locally installed executables? Or is this something against `atom-beautify` philosophy?

szeck87:
The closest thing I can find is cljfmt and it's completely out of date (using `run` instead of `exe`).  It doesn't use a global, just a local.

kachkaev:
Maybe we could use a global-only instance of remark for now then. However, may be a good practice to search for an executable in project deps first and give that instance a preference. Have seen in a couple of places; [prettier-atom](https://atom.io/packages/prettier-atom) is on top of my head.

szeck87:
Where in prettier-atom specifically?

kachkaev:
I did not inspect their source, yet, but this conclusion comes from my experience and from these options in preferences:

szeck87:
That just looks like it won't format if you don't have prettier as a dependency in the directory of where you're working.

Anyhow, like I mentioned earlier, ideally we don't use executables unless there is no other alternative.  If you can find a way to get the cli version to run within node, that would be great.  Otherwise I'll defer . to @Glavin001 on what to do.

wooorm:
@kachkaev 

> In any case, turning .remarkrc statements like [require('remark-lint-heading-style'), 'setext'] or [require('remark-lint-ordered-list-marker-value'), 'one'] into some kind of eslint --fix still feels like a missing feature.

Agreed! Related to remarkjs/remark-lint#82!

Glavin001:
> Anyhow, like I mentioned earlier, ideally we don't use executables unless there is no other alternative. If you can find a way to get the cli version to run within node, that would be great. 

I would agree with @szeck87 .

> .remarkrc files are only supported by remark-cli. See remark-cli for more details!

However, in this case using `remark-cli` may be required to support `.remarkrc`.

This route is acceptable as long as users still have the option to configure via `.jsbeautifyrc` or Atom-Beautify package [settings view](https://github.com/atom/settings-view).

package.json
```diff
@@ -185,7 +185,6 @@
     "open": "0.0.5",
     "prettydiff2": "^2.2.7",
     "pug-beautify": "^0.1.1",
-    "remark": "6.0.1",
```
Glavin001: @kachkaev @szeck87 : I think we can support both Remark-CLI and be backwards compatible with `remark` as a Node.js dependency.

Since the `Remark CLI` is now an Executable we should be able to determine if it is installed or not.
See https://github.com/Glavin001/atom-beautify/blob/30dece209a50868c5ce8589b69a79240e91e5ea4/src/beautifiers/executable.coffee#L99-L107

I'm thinking something along the lines of:

```coffeescript
isCliInstalled = @exe("remark").isInstalled

if isCliInstalled
  # ...
  @exe("remark").run(remarkArgs...
  # ...
else
  # ...
  cleanMarkdown = remark().process(text, options).toString()	
  # ...
```

I would prefer to add new features while being backwards compatible. Removable of `remark` as a builtin dependency is considered a breaking change to me. However, there is value in using `Remark CLI` directly 👍, such as the configuration file.

kachkaev:
@Glavin001 will you have time to finish these changes by a chance? I'm afraid I'll be quite busy for the next several weeks. Feel free to take over this PR and merge it! And thanks for your review!

szeck87:
@kachkaev do you still want to move forward with this, or just use Prettier combined with your other PR for formatting code blocks in markdown?

kachkaev:
Hi @szeck87! I have switched to Prettier and am currently betting on its use with [plugins](https://prettier.io/docs/en/plugins.html). I won't be able to finish this PR but you are free to continue with its implementation.

szeck87:
OK no problem, I'll close this out.

PR #2009, opened 2018-01-12
Fallback to global eslint and add errors_only option

### LINKED ISSUE #1739: Add support for globally installed ESLint fixer
# Description

I got an error when trying to beautify when ESLint is not installed in the local project.

# Input Before Beautification

# Steps to Reproduce

1. Install ESLint globally
2. Add code to Atom editor
3. Use the ESLint Fixer for Javascript
4. Run command `Atom Beautify: Beautify Editor`
5. Got an error that ESLint is not found

# Debug

Here is a link to the `debug.md` Gist: <INSERT GIST HERE>

# Checklist

I have:
1. Installed ESLint on the local project and it work.

*(Depending on the extension you use in the .eslintrc you will need to install them)

UziTech:

checks for a local eslint but falls back to the global eslint

```
importPath = Path.join(projectPath, 'node_modules', 'eslint')
try
  eslint = require(importPath)
catch
  eslint = require('eslint')
```

Adds an option to only fix errors and leave warnings alone.

fixes #1739

### Checklist

### DIFF
--- a/src/beautifiers/eslint.coffee
+++ b/src/beautifiers/eslint.coffee
@@ -9,7 +9,8 @@ module.exports = class ESLintFixer extends Beautifier
   link: "https://github.com/eslint/eslint"
 
   options: {
-    JavaScript: false
+    JavaScript:
+      errors_only: true
     Vue: false
   }
 
@@ -23,9 +24,16 @@ module.exports = class ESLintFixer extends Beautifier
       allowUnsafeNewFunction ->
         importPath = Path.join(projectPath, 'node_modules', 'eslint')
         try
-          CLIEngine = require(importPath).CLIEngine
+          try
+            eslint = require(importPath)
+          catch
+            eslint = require('eslint')
 
-          cli = new CLIEngine(fix: true, cwd: projectPath)
+          fix = true
+          if options.errors_only
+            fix = (rule) -> rule.severity is 2
+
+          cli = new eslint.CLIEngine(fix: fix, cwd: projectPath)
           result = cli.executeOnText(text).results[0]
 
           resolve result.output
--- a/src/beautifiers/eslint.coffee
+++ b/src/beautifiers/eslint.coffee
@@ -18,26 +18,41 @@ module.exports = class ESLintFixer extends Beautifier
     return new @Promise((resolve, reject) ->
       editor = atom.workspace.getActiveTextEditor()
       filePath = editor.getPath()
-      dir = path.dirname(filePath)
-      projectPath = atom.project.relativizePath(filePath)[0]
+
+      if filePath
+        dir = path.dirname(filePath)
+        projectPath = atom.project.relativizePath(filePath)[0]
+      else
+        dir = atom.getConfigDirPath()
+        projectPath = atom.project.getPaths()[0]
 
       result = null
+      importPath = Path.join(projectPath or dir, 'node_modules', 'eslint')
+      try
+        eslint = require(importPath)
+      catch
+        eslint = require('eslint')
+
+      fix = true
+      if options.errors_only
+        fix = (rule) -> rule.severity is 2
+
+      eslintExecute = (cwd) ->
+        cli = new eslint.CLIEngine(fix: fix, cwd: cwd)
+        result = cli.executeOnText(text).results[0]
+        result.output
+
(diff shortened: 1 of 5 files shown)


AbdelrahmanHafez:
Any news on this?

UziTech:
@Glavin001 feedback would be nice

UziTech:
Looks like the travis-ci build failed because of configuration

src/beautifiers/eslint.coffee
```diff
@@ -9,25 +9,49 @@ module.exports = class ESLintFixer extends Beautifier
...
       result = null
-      allowUnsafeNewFunction ->
-        importPath = Path.join(projectPath, 'node_modules', 'eslint')
-        try
-          CLIEngine = require(importPath).CLIEngine
+      importPath = Path.join(projectPath or dir, 'node_modules', 'eslint')
+      try
+        eslint = require(importPath)
+      catch
+        eslint = require('eslint')
```
Glavin001: I don't think `require('eslint')` is sufficient to be "global". By global, you mean installed via `npm install --global eslint`?
See https://stackoverflow.com/a/15646750/2578205

Glavin001: This would work if `eslint` was a dependency in Atom-Beautify's `package.json` file.

UziTech: eslint is dependency in Atom-Beautify's package.json file.

Glavin001: Correct: https://github.com/Glavin001/atom-beautify/blob/master/package.json#L191

However I want to confirm the goal here. Is it to use the `eslint` installed by `npm install --global eslint` (global) or the one installed within Atom-Beautify itself? The latter is what is being done currently, however the description makes me think the former is what you actually want.
We could support all three options, however something like [`requireg`](https://www.npmjs.com/package/requireg) will be required.

UziTech: I meant global as just not local to the project. Using the eslint included with Atom-Beautify is fine with me.

UziTech: I could check for *global* eslint before the included one if that is what people want

something like:

```coffeescript
try
  eslint = require(importPath)
catch
  try
    eslint = requireg('eslint')
  catch
    eslint = require('eslint')
```

Glavin001: This looks like a good idea: future proof. 👍

src/options.json
```diff
@@ -1340,6 +1340,20 @@
           "namespace": "js"
         }
       },
+      "errors_only": {
+        "type": "boolean",
+        "default": false,
+        "description": "Fix errors but leave warnings (Supported by JS Beautify)",
+        "title": "Errors only",
+        "beautifiers": [
+          "JS Beautify"
```
Glavin001: Looks like the docs need updating. `npm run docs`

src/languages/javascript.coffee
```diff
@@ -112,5 +112,9 @@ module.exports = {
       type: 'boolean'
       default: false
       description: "Insert spaces between brackets in object literals"
+    errors_only:
+      type: 'boolean'
+      default: false
+      description: "Fix errors but leave warnings"
```
Glavin001: Can we change the message to be more specific and formal:

```
Fix ESLint errors and exclude warnings
```

src/beautifiers/eslint.coffee
UziTech: looks like `requireg` checks the regular `require` first so it would always grab the included one 😕 

We could just add an eslint option to the Executables settings so people could specify there own executable if they wanted to.

I'm not sure how to add a new option under Executables

src/options.json
UziTech: right now JS Beautify is set to use all javascript options. how do I tell it to use all except error_only?

szeck87:
This will be fixed natively once Atom Beautify uses Unibeautify completely.

szeck87:
@UziTech Thanks for the update.  I'll close this PR, but if anyone wants to pick it up they're welcome to.

Glavin001:
For those subscribed, global ESLint support is being continued with https://github.com/Glavin001/atom-beautify/pull/2147/files

PR #2010, opened 2018-01-15
Introduce internal flag `textIsNotEntireFile` inside `context` argument for `beautify()`

kachkaev:
This should help some beautifiers such as `elm-formatter` correctly beautify file fragments in future (e.g. avoid adding unexpected imports, extra new lines, etc.)

When a beautifier formats something, it is not aware if the given text represents the entire file or just a fraction of it. A new flag called `textIsNotEntireFile` has been introduced to fix this. This flag is now passed to all `beautify()` methods as a part of the fourth argument (`context`). It is the beautifier's choice what to do with this (nothing needs to be done by default).

Not sure, because I haven't done a rigorous search.

### Any other comments?

I'd like to use `textIsNotEntireFile` with Elm as soon as https://github.com/avh4/elm-format/issues/65 is fixed and so passing a flag to `elm-format` makes sense. This new `textIsNotEntireFile` flag may be also a handy addition to https://github.com/Glavin001/atom-beautify/pull/1990.

### Checklist

- [ ] <s>Added examples for testing to [examples/ directory](examples/)</s>

---

PS: Where should I document the presence of this new flag? Could not find any breakdown of what `context` argument contains.

### DIFF
--- a/src/beautify.coffee
+++ b/src/beautify.coffee
@@ -168,11 +168,14 @@ beautify = ({ editor, onSave, language }) ->
 
     # Get current editor's text
     text = undefined
+    textIsNotEntireFile = false
     if not forceEntireFile and isSelection
       text = editor.getSelectedText()
+      textIsNotEntireFile = editor.getSelectedText().length < editor.getText().length
     else
       text = editor.getText()
     oldText = text
+    logger.verbose('textIsNotEntireFile', textIsNotEntireFile)
 
 
     # Get Grammar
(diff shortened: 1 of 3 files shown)


kachkaev:
WDYT @szeck87?

src/beautify.coffee
```diff
@@ -168,11 +168,14 @@ beautify = ({ editor, onSave, language }) ->
 
     # Get current editor's text
     text = undefined
+    textIsNotEntireFile = false
     if not forceEntireFile and isSelection
       text = editor.getSelectedText()
+      textIsNotEntireFile = editor.getSelectedText().length < editor.getText().length
```
Glavin001: How is `isSelection` different from the intention of `textIsNotEntireFile`?

kachkaev: You can select the entire file and beautify it. In this case you don't want the selection to be treated the same as any arbitrary code chunk.

For instance, if one day we have https://github.com/avh4/elm-format/issues/65 resolved, `isSelection` will work incorrectly for Elm. In particular, when selecting the entire file and beautifying it people would expect to see an automatically added module definition line, but this would not happen. Introducing `textIsNotEntireFile` solves this potential issue in Elm and in other languages.

kachkaev:
Closing this PR as stale. @Glavin001 feel free to reopen it or recycle the code 🙌

PR #1926, opened 2017-11-03
Added option for sqlformat to re-indent

faheel:
### What does this implement/fix? Explain your changes.

Added an option to either re-indent code or to preserve indentation. The SQL formatter, sqlparse, has an option for exactly this and is useful when the indentation is to be left unchanged.

### Does this close any currently open issues?

No.

### Any other comments?

...

### Checklist

Check all those that are applicable and complete.

- [x] Merged with latest `master` branch
- [x] Regenerate documentation with `npm run docs`
- [x] Add change details to `CHANGELOG.md` under "Next" section
- [ ] Added examples for testing to [examples/ directory](examples/)
- [x] Travis CI passes (Mac support)
- [x] AppVeyor passes (Windows support)

PR #2064, opened 2018-03-03
Fix "Reindent" option for sqlformat.

pengjunjie1207:

For sqlparse-0.2.3 and sqlparse-0.2.4, beautifying sql will throw "sqlformat: error: argument -r/--reindent: ignored explicit argument". I remove the param asignment in the argument and change if condition.

No.

### Any other comments?

The code works with user's config about "Reindent" in settings->atom-beautify->sql and I hava tested it in both Ubuntu and Windows.

### Checklist

It is the first time I use Travis CI. It passed linux and failed osx, saying "Couldn't set selectedTextBackgroundColor".

### DIFF
--- a/src/beautifiers/sqlformat.coffee
+++ b/src/beautifiers/sqlformat.coffee
@@ -17,7 +17,7 @@ module.exports = class Sqlformat extends Beautifier
   beautify: (text, language, options) ->
     @run("sqlformat", [
       @tempFile("input", text)
-      "--reindent=#{options.reindent}" if options.reindent?
+      "--reindent" if (options.reindent? && options.reindent == true)
       "--indent_width=#{options.indent_size}" if options.indent_size?
       "--keywords=#{options.keywords}" if (options.keywords? && options.keywords != 'unchanged')
       "--identifiers=#{options.identifiers}" if (options.identifiers? && options.identifiers != 'unchanged')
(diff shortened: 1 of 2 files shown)


Glavin001:
Related to https://github.com/Glavin001/atom-beautify/pull/1926

src/beautifiers/sqlformat.coffee
```diff
@@ -17,7 +17,7 @@ module.exports = class Sqlformat extends Beautifier
   beautify: (text, language, options) ->
     @run("sqlformat", [
       @tempFile("input", text)
-      "--reindent=#{options.reindent}" if options.reindent?
+      "--reindent" if (options.reindent? && options.reindent == true)
```
Glavin001: I think this should be:

```coffeescript
"--reindent" if options.reindent is true
```

Which is the following in JavaScript:

```javascript
if (options.reindent === true) {
  "--reindent";
}
```

pengjunjie1207: What about the situation that options.reindent is null or undefined? Or it is impossible to meet that situation.

Glavin001: It could happen and then it would not add `--reindent` option at all. Is this the desired behaviour?

Glavin001:
@szeck87 Do you recall why the SQL test case was disabled?
See https://github.com/Glavin001/atom-beautify/tree/master/examples/nested-jsbeautifyrc/sql/original
We have `sqlparse` installed for Travis CI: https://github.com/Glavin001/atom-beautify/blob/master/requirements.txt#L3

Glavin001:
Published to v0.32.2

PR #2065, opened 2018-03-04
Unibeautify

szeck87:
* Updates dependencies and package scripts
* Script to create Atom Settings includes beautifiers to include for a language (currently an array of strings unfortunately)
* Beautifying passes language options
...

...

### Any other comments?

...

### Checklist


script/index.ts
```diff
@@ -63,6 +63,13 @@ function buildOptionsForLanguage(language: Language, beautifiers: String[]) {
       type: "string"
     }
   }
+  languageOptions["beautify_on_save"] = {
```
Glavin001: Thinking about this more and more and I feel like these kind of decisions should be User and/or Editor specific. Storing in a file means this is a project level concern. While selected beautifiers and options make sense to project level, enforcing WHEN I beautify such as `beautify_on_save` feels like the User should be personally setting this in their editor. For instance, I do NOT use beautify on save. I run it manually when I want it. However, looking at the stats, most people prefer the beautify on save option to be enabled.

Thoughts? I do not want to remove this from PR, just discuss it :).

szeck87: It is for the editor, it's the script that generates the options.json file for Atom Beautify's settings.

script/index.ts
```diff
@@ -37,8 +37,8 @@ function buildOptions() {
   return options;
 }
 
-function buildOptionsForLanguage(language: Language) {
-  let languageOptions = {};
+function buildOptionsForLanguage(language: Language, beautifiers: String[]) {
+  let languageOptions: any = {};
```
Glavin001: Should be a more concrete interface. Do the types for Atom have any config related interfaces?

script/index.ts
```diff
@@ -55,14 +55,29 @@ function buildOptionsForLanguage(language: Language) {
       items: option.items
     };
   })
+  languageOptions["Beautifiers"] = {
+    title: "Beautifiers",
+    type: "array",
+    default: beautifiers,
+    items: {
+      type: "string"
```
Glavin001: Could this be an enum from `beautifiers`? Basically, `beautifiers` option is an array of beautifier names, in order, and can pick any of the beautifiers names (not all `string`s).

szeck87:
There is a lot of duplicate code in AtomBeautify.ts that I need to clean up between the `beautifyOnSaveHandler` and `beautifyEditor` methods.  Any help there would be appreciated.

src/config.ts
```diff
@@ -29,7 +29,7 @@ export default {
                 title: "Beautify On Save",
                 type: "boolean",
                 "default": true,
-                description: "When saving a file, the contents will be beautified."
+                description: "When saving a file, the contents will be beautified (unchecking will override individual language setting)."
```
Glavin001: `beautify_on_save` should be language specific. Remove from global.

script/index.ts
Glavin001: Nevermind. I thought this was for extracting from the configuration file.

Although I do want to have a discussion in a separate thread about the above and what options should be supported in the configuration file.

script/index.ts
szeck87: Not from what I can tell, https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/atom/index.d.ts.

script/index.ts
szeck87: Not possible with Atom's settings UI unfortunately.

script/index.ts
Glavin001: Lame. Let's create our own interface, at least. Something simple.

szeck87:
@Glavin001 made the changes requested.

script/index.ts
```diff
@@ -55,16 +55,56 @@ function buildOptionsForLanguage(language: Language) {
       items: option.items
     };
   })
+  languageOptions["Beautifiers"] = {
```
Glavin001: Shouldn't this be `beautifiers` instead of `Beautifiers`?

https://github.com/Unibeautify/unibeautify/blob/4f5a4b442ecf173c498435a5d0d16aa0499c5e5d/test/beautifier/pipeline.spec.ts#L13

src/AtomBeautify.ts
```diff
@@ -3,10 +3,10 @@ import * as Atom from "atom";
 import beautifiers from "./beautifiers";
 import { CompositeDisposable, Disposable } from "atom";
 import Config from "./config";
-import * as Promise from "bluebird";
 import * as path from "path";
 import * as _ from "lodash";
 const pkg = require("../package");
+const cosmiconfig = require("cosmiconfig");
```
Glavin001: Should be using `import`.

`@types/cosmiconfig` exists: https://yarnpkg.com/en/package/@types/cosmiconfig

src/AtomBeautify.ts
```diff
@@ -85,13 +93,23 @@ export class AtomBeautify {
...
+      const language = langs.length > 0 ? langs[0] : null;
+      const configSettings = this.getConfigFromSettings(language.name);
+      const beautifySettings = await this.unibeautifyConfiguration(configSettings, editorPath, language.name);
       return this.unibeautify.beautify({
         fileExtension,
         atomGrammar: grammarName,
-        options: {},
+        options: {
+          [language.name]: beautifySettings
+        },
```
Glavin001: Should be passing in the configs for all languages. A single `beautify` could end up beautifying multiple languages. Consider `PHP` as root language and having embedded `HTML`.

script/index.ts
szeck87: Hmm, I thought I saw in the interface it was capitalized when that was committed. But yea, you're right:  https://github.com/Unibeautify/unibeautify/blob/master/src/beautifier.ts#L308

src/AtomBeautify.ts
szeck87: I agree.  At the same time that feature warrants more discussion.

src/AtomBeautify.ts
Glavin001: I ended up needing to add the typings myself. Appears @types/cosmiconfig may be broken.

- https://github.com/Unibeautify/vscode/blob/master/src/%40types/cosmiconfig/index.d.ts
- https://github.com/Unibeautify/vscode/blob/master/tsconfig.json#L13-L19

src/AtomBeautify.ts
Glavin001: Here's what I did for VSCode:
- https://github.com/Unibeautify/vscode/blob/master/src/EditProvider.ts#L102-L115
- https://github.com/Unibeautify/vscode/blob/master/src/EditProvider.ts#L60-L68

src/AtomBeautify.ts
```diff
@@ -130,9 +114,9 @@ export class AtomBeautify {
       // TODO
     }
 
-    private getConfigFromSettings(language: string) {
+    private configFromSettings(language?: Language) {
       const config = atom.config.get("atom-beautify");
-      return _.get(config, language.toLowerCase());
+      return language ? _.get(config, language.name) : config;
```
Glavin001: I think `config` and `config[language.name]` are very different types.

For example:

```
const config = {
  "JavaScript": {
    "indent_size": 2
  }
};
```

PR #2118, opened 2018-04-21
Add Smarty language

brunoAge:
Smarty is a template engine for PHP

### DIFF
--- a/src/languages/index.coffee
+++ b/src/languages/index.coffee
@@ -64,6 +64,7 @@ module.exports = class Languages
     "rust"
     "sass"
     "scss"
+    "smarty"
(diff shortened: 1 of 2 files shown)


Bruno-basswn:
I know Smarty does not have the same HTML structure, but it would already be something.

szeck87:
@Bruno-basswn please add examples in the `examples` directory, and also run `npm run docs` and push again.

JFATech:
For me, the changes in the two files are not working. I still get the warning "could not find a supported beautifier". Grammar "Smarty" and extension "tpl".
Is there any hope to get Smarty supported anytime soon?

PR #2147, opened 2018-05-30
eslint beautifier bugfix

zhuyingda:

Eslint beautifier should load globally eslint module which install by `npm i -g eslint` if it do installed.
And there should be some warning to users when it got some problem.

In this project, no.
But I think there should be some individuals have this problem.

Let me know if any code or desc I commited doesn't match this project rules.
Thank you.

### DIFF
--- a/src/beautifiers/eslint.coffee
+++ b/src/beautifiers/eslint.coffee
@@ -14,6 +14,7 @@ module.exports = class ESLintFixer extends Beautifier
   }
 
   beautify: (text, language, options) ->
+    context = this
     return new @Promise((resolve, reject) ->
       editor = atom.workspace.getActiveTextEditor()
       filePath = editor.getPath()
@@ -23,12 +24,25 @@ module.exports = class ESLintFixer extends Beautifier
       allowUnsafeNewFunction ->
         importPath = Path.join(projectPath, 'node_modules', 'eslint')
         try
-          CLIEngine = require(importPath).CLIEngine
+          context.which('eslint').then((globalEslintPath) ->
+            if globalEslintPath != 'eslint'
+              importPath = Path.resolve(globalEslintPath, '../../lib/node_modules/eslint')
+            else
+              needInstallMsg = 'Error: it seems like you did not install eslint globally,'
+              needInstallMsg += 'try to run npm i -g eslint and restart the atom editor.'
+              context.deprecate needInstallMsg
+              resolve undefined
+            CLIEngine = require(importPath).CLIEngine
 
-          cli = new CLIEngine(fix: true, cwd: projectPath)
-          result = cli.executeOnText(text).results[0]
-
-          resolve result.output
+            cli = new CLIEngine(fix: true, cwd: projectPath)
+            result = cli.executeOnText(text).results[0]
+            if not result.output
+              errMsg = result.messages[0].message
+              errLine = result.messages[0].line
+              errColumn = result.messages[0].column
+              context.deprecate 'Error: [' + errMsg + '] at line ' + errLine + ', column ' + errColumn
+            resolve result.output
+          )
         catch err
           reject(err)
     )
--- a/src/beautifiers/eslint.coffee
+++ b/src/beautifiers/eslint.coffee
@@ -3,6 +3,7 @@
 Beautifier = require('./beautifier')
 Path = require('path')
 {allowUnsafeNewFunction} = require 'loophole'
+requireg = require('requireg');
 
 module.exports = class ESLintFixer extends Beautifier
   name: "ESLint Fixer"
@@ -24,10 +25,10 @@ module.exports = class ESLintFixer extends Beautifier
         importPath = Path.join(projectPath, 'node_modules', 'eslint')
         eslint = null
         try
-          eslint = require(importPath)
+          eslint = requireg('eslint')
         catch
           try
-            eslint = requireg('eslint')
+            eslint = require(importPath)
           catch
             try
               eslint = require('eslint')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Next
 
+# v0.32.5 (2018-05-28)
+- Fixes an issue with Rubocop not working on Windows ([#2092](https://github.com/Glavin001/atom-beautify/issues/2092))
+- Fixes an issue with PHPCBF not running when specifying a path to it ([#2140](https://github.com/Glavin001/atom-beautify/issues/2140))
+- Fixes the repository link to Atom Beautify within Atom's settings ([#2086](https://github.com/Glavin001/atom-beautify/pull/2086))
+
 # v0.32.4 (2018-05-15)
 - Fix php-cs-fixer not running when using `.phar` files ([#2134](https://github.com/Glavin001/atom-beautify/pull/2134))
 - Fix error `Error is not defined` ([#2134](https://github.com/Glavin001/atom-beautify/pull/2134))
--- a/package.json
+++ b/package.json
@@ -204,6 +204,7 @@
     "prettydiff2": "^2.2.7",
     "pug-beautify": "^0.1.1",
     "remark": "6.0.1",
+    "requireg": "^0.1.8",
     "season": "^6.0.2",
     "semver": "^5.5.0",
     "shell-env": "^0.3.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "atom-beautify",
   "main": "./src/beautify",
-  "version": "0.32.4",
+  "version": "0.32.5",
   "private": true,
   "description": "Beautify HTML, CSS, JavaScript, PHP, Python, Ruby, Java, C, C++, C#, Objective-C, CoffeeScript, TypeScript, Coldfusion, SQL, and more in Atom",
   "repository": "https://github.com/Glavin001/atom-beautify.git",


src/beautifiers/eslint.coffee
```diff
@@ -23,12 +24,23 @@ module.exports = class ESLintFixer extends Beautifier
       allowUnsafeNewFunction ->
         importPath = Path.join(projectPath, 'node_modules', 'eslint')
         try
-          CLIEngine = require(importPath).CLIEngine
+          importPath = Path.join(projectPath, 'node_modules', 'eslint')
+          context.which('eslint').then((globalEslintPath) ->
+            if `globalEslintPath !== 'eslint'`
```
stevenzeck: Please use double and single quotes, not backticks.

src/beautifiers/eslint.coffee
```diff
@@ -23,12 +24,23 @@ module.exports = class ESLintFixer extends Beautifier
       allowUnsafeNewFunction ->
         importPath = Path.join(projectPath, 'node_modules', 'eslint')
         try
-          CLIEngine = require(importPath).CLIEngine
+          importPath = Path.join(projectPath, 'node_modules', 'eslint')
```
stevenzeck: We shouldn't need `importPath` here as its initialized the same way two lines above it.

src/beautifiers/eslint.coffee
```diff
@@ -23,12 +24,23 @@ module.exports = class ESLintFixer extends Beautifier
...
+              context.deprecate 'Error: it seems like you did not install eslint globally, try to run npm i -g eslint'
+            CLIEngine = require(importPath).CLIEngine
 
-          cli = new CLIEngine(fix: true, cwd: projectPath)
-          result = cli.executeOnText(text).results[0]
-
-          resolve result.output
+            cli = new CLIEngine(fix: true, cwd: projectPath)
+            result = cli.executeOnText(text).results[0]
+            if not result.output
```
stevenzeck: What does this block do exactly?  From Line 37 to 41.

zhuyingda: Here is the case: There is a `.eslintrc` config file in the `projectPath`,  and there are some `async/await` syntax in the current js code which input by `text` param. However, I did not set `ecmaVersion: 8` in my `.eslintrc` so that the `result` of `cli.executeOnText(text).results[0]` will get `undefined`.

> Reference:  [eslint project issue](https://github.com/eslint/eslint/issues/8126)

In the current logic of eslint-beautifier from the master branch, there is not any warning or error notification for user, it just get nothing. So these logic is a friendly error message for user.

src/beautifiers/eslint.coffee
zhuyingda: Okay, I change this in my later commit

src/beautifiers/eslint.coffee
zhuyingda: Yes, sorry for that, I will delete this on my next fixed commit.

zhuyingda:
Hi, @szechyjs I have update my pr by commit some fixed code.
Please review again. 😊

szechyjs:
@szeck87 see above

zhuyingda:
@szeck87 
I have update my pr by commit some fixed code.
Please review again.

src/beautifiers/eslint.coffee
```diff
@@ -23,12 +24,25 @@ module.exports = class ESLintFixer extends Beautifier
...
         importPath = Path.join(projectPath, 'node_modules', 'eslint')
         try
-          CLIEngine = require(importPath).CLIEngine
+          context.which('eslint').then((globalEslintPath) ->
+            if globalEslintPath != 'eslint'
+              importPath = Path.resolve(globalEslintPath, '../../lib/node_modules/eslint')
+            else
+              needInstallMsg = 'Error: it seems like you did not install eslint globally,'
+              needInstallMsg += 'try to run npm i -g eslint and restart the atom editor.'
+              context.deprecate needInstallMsg
```
Glavin001: This message looks like it should be an error message to the user, not a deprecation notice.

src/beautifiers/eslint.coffee
```diff
@@ -14,6 +14,7 @@ module.exports = class ESLintFixer extends Beautifier
   }
 
   beautify: (text, language, options) ->
+    context = this
```
Glavin001: I do not think this self referential is necessary.
See https://stackoverflow.com/a/8965984/2578205 and https://coffeescript.org/#fat-arrow

Alternatively:

Before:

```coffeescript
context = this
# .... 
context.which('eslint').then(...)
```

After:

```coffeescript
eslint = this.which('eslint')
# .... 
eslint.then(...)
```

Also note `this` should be replaced with `@` in CoffeeScript. See https://coffeescript.org/#operators

src/beautifiers/eslint.coffee
```diff
@@ -23,12 +24,25 @@ module.exports = class ESLintFixer extends Beautifier
       allowUnsafeNewFunction ->
         importPath = Path.join(projectPath, 'node_modules', 'eslint')
         try
-          CLIEngine = require(importPath).CLIEngine
+          context.which('eslint').then((globalEslintPath) ->
+            if globalEslintPath != 'eslint'
+              importPath = Path.resolve(globalEslintPath, '../../lib/node_modules/eslint')
```
Glavin001: The usage of `context.which('eslint')` and the `../../lib/node_modules/eslint` concerns me.

I would prefer something more formal and less fragile.

[`requireg`](https://www.npmjs.com/package/requireg) can be used to require global modules.

See https://github.com/Glavin001/atom-beautify/pull/2009/files#r171870409 for what I'd like to see:

```coffeescript
try
  eslint = require(importPath) # Local ESLint for project
catch
  try
    eslint = requireg('eslint') # Globally installed ESLint
  catch
    eslint = require('eslint') # ESLint packaged with Atom-Beautify
```

src/beautifiers/eslint.coffee
```diff
@@ -23,12 +24,25 @@ module.exports = class ESLintFixer extends Beautifier
...
-          result = cli.executeOnText(text).results[0]
-
-          resolve result.output
+            cli = new CLIEngine(fix: true, cwd: projectPath)
+            result = cli.executeOnText(text).results[0]
+            if not result.output
+              errMsg = result.messages[0].message
+              errLine = result.messages[0].line
+              errColumn = result.messages[0].column
+              context.deprecate 'Error: [' + errMsg + '] at line ' + errLine + ', column ' + errColumn
```
Glavin001: `deprecate` does not appear to be applicable here. Please `reject(new Error("insert message here"))` and/or use the logger methods (e.g. `      @verbose("insert message here")`)

Glavin001: What does `result.messages[0]` look like? If it is an `Error` instance than please simply:

```coffeescript
error = result.messages[0]
return reject(error)
```

src/beautifiers/eslint.coffee
zhuyingda: Yeah, it is better.
I will change the code and update my pull request.

src/beautifiers/eslint.coffee
zhuyingda: That is really a good way, I will try to change the method and test if that is okay on my mac os environment.

src/beautifiers/eslint.coffee
zhuyingda: Understand.
Can I use `@verbose("error notification message")` here instead of `deprecation`?

Glavin001: Yes, `@verbose(needInstallMsg)` should work. Let me know if you have any problems.

src/beautifiers/eslint.coffee
UziTech: This won't work. requireg only searches for a global install if `require(...)` fails.

See [requireg priority](https://www.npmjs.com/package/requireg#resolution-priority)

zhuyingda:
@Glavin001 
I have updated my pull request, without `which('eslint')` but as your recommended way.
Now the beautifier will require eslint module as this sequence:
1. current project path eslint module
2. global eslint
3. atom-beautifier package.json eslint
Otherwise it would show the "not installed message to user", but it seems like it will never show this, unless atom-beautifier delete eslint in package.json.

Please review again.

src/beautifiers/eslint.coffee
```diff
@@ -14,35 +14,35 @@ module.exports = class ESLintFixer extends Beautifier
...
+        eslint = null
         try
-          context.which('eslint').then((globalEslintPath) ->
-            if globalEslintPath != 'eslint'
-              importPath = Path.resolve(globalEslintPath, '../../lib/node_modules/eslint')
-            else
+          eslint = require(importPath)
+        catch
+          try
+            eslint = requireg('eslint')
```
UziTech: `requireg('eslint')` won't work here since it just calls `require('eslint')` first and only checks for a global eslint if none is provided with atom-beautify

zhuyingda: Thanks, I have installed the `requireg` package to atom-beautify.
Plz see my newest commit.

UziTech: How is installing requireg going to change is priority?

UziTech:
`requireg('eslint')` won't work here since requireg just calls `require('eslint')` first and only checks for a global eslint if none is provided with atom-beautify

src/beautifiers/eslint.coffee
zhuyingda: I have change the priority in the newest update pr.
Plz check again.

zhuyingda:
Thanks to @UziTech helps.
I have change the priority of `requireg` and `require`, and installed the `requireg` npm package to atom-beautify.

UziTech:
@zhuyingda I don't think you understand what I'm saying. `requireg` just calls `require` so the global eslint will **NEVER** be called if you use `requireg`

zhuyingda:
@UziTech Okay, I know.
So the only way solving this problem is to remove `eslint` package in atom-beautify, is there any other way?

UziTech:
Not sure if there is another way other than manually checking wherever `requireg` checks after checking`require`

zhuyingda:
If so, perhaps we should remove eslint package in atom-beautify `node-modules` and `package.json` so that the `requireg` would works fine.
Is that okay for the whole project? @Glavin001 
Plz see this.

UziTech:
@zhuyingda Why would someone rather use global eslint over the included one?

zhuyingda:
> @zhuyingda Why would someone rather use global eslint over the included one?

@UziTech That is a quite reasonable case.
Sometimes the `eslint` or other formatter plugin for `js` or for other languages has a higher version than the same formatter package in atom-beautify, people would use global ones instead.
Also, I think it is a coincidence that there is a `eslint` package in atom-beautify directory because atom is a js-family (electron, coffeescript) editor. For other formatter, you have to install their package globally because there is no package in atom-beautify directiory, such as php-cs-fixer(a php formatter), autopep8(a python formatter), etc.

src/beautifiers/eslint.coffee
Glavin001: Instead of `requireg` looks like `import-global` should work: https://github.com/sindresorhus/import-global

src/beautifiers/eslint.coffee
Glavin001: > requireg('eslint') won't work here since it just calls require('eslint') first and only checks for a global eslint if none is provided with atom-beautify

Instead of `requireg` looks like `import-global` should work: https://github.com/sindresorhus/import-global

src/beautifiers/eslint.coffee
```diff
@@ -14,21 +15,35 @@ module.exports = class ESLintFixer extends Beautifier
...
         try
-          CLIEngine = require(importPath).CLIEngine
-
-          cli = new CLIEngine(fix: true, cwd: projectPath)
-          result = cli.executeOnText(text).results[0]
-
-          resolve result.output
-        catch err
-          reject(err)
+          eslint = requireg('eslint')
```
Glavin001: This order is incorrect.
We want:
1. Project's locally installed ESLint
2. Globally installed ESLint
3. Atom-Beautify's fallback ESLint

As @UziTech [pointed out](https://github.com/Glavin001/atom-beautify/pull/2147#discussion_r192610452), `requireg` is not actually what we want.

I think https://github.com/sindresorhus/import-global is what we want.

```coffeescript
try
  importPath = Path.join(projectPath, 'node_modules', 'eslint')
  eslint = require(importPath) # Local ESLint for project
catch
  try
    importGlobal = require('import-global');
    eslint = importGlobal('eslint') # Globally installed ESLint
  catch
    eslint = require('eslint') # ESLint packaged with Atom-Beautify
```

zhuyingda: It seems like `import-global` package will find the module by `/Applications/Atom.app/Contents/Frameworks/Atom Helper.app/Contents/lib/node_modules/eslint` like this.
I installed the `eslint` globally by npm and its path is in `xx/xx/lib/node_modules/eslint` where the npm bin file is installed.

Should I choose the `which` method in atom-beautify instead of importGlobal?

Glavin001: @zhuyingda `import-global` also did not work? Wow.

Although idea is using `requireg` and manually calling the internal `nodePathResolve` function: https://github.com/h2non/requireg/blob/master/lib/resolvers.js#L37-L54

```javascript
const nodePathResolve = require("requireg/lib/resolvers")[1];
const globalPackage = nodePathResolve(module, "eslint");
```

Until @UziTech 's PR is merged: https://github.com/h2non/requireg/pull/15

zhuyingda:
It is still a issue, please dont mark it.

src/beautifiers/eslint.coffee
UziTech: https://github.com/h2non/requireg/pull/15 is merged in [requireg v0.2.0](https://www.npmjs.com/package/requireg)

so `eslint = requireg('eslint', true)` should work to install globally

zhuyingda:
It is still a issue, please dont mark it.

src/beautifiers/eslint.coffee
Glavin001: @zhuyingda Looks like this should be changed:

```suggestion
          eslint = requireg('eslint', true)
```

zhuyingda:
@Glavin001 It seems like this issue has been waiting for merging for months, is there any plan to solve this issue now? Really appreciate it.

src/beautifiers/eslint.coffee
```diff
@@ -14,21 +15,35 @@ module.exports = class ESLintFixer extends Beautifier
...
-          reject(err)
+          eslint = requireg('eslint', true)
+        catch
+          try
+            eslint = require(importPath)
+          catch
+            try
+              eslint = require('eslint')
+            catch
+              needInstallMsg = 'Error: it seems like you did not install eslint globally,'
```
Glavin001, line 36: The current recommended way to install ESLint is not globally: https://eslint.org/docs/user-guide/getting-started#installation-and-usage

This leads me to the conclusion the order of line 28 to line 34 should be changed:

1) `require(importPath` - Project specific
2) `requireg('eslint', true)` - Globally
3) `require('eslint')` - Atom-Beautify

[Since Atom-Beautify has it installed as a direct dependency](https://github.com/Glavin001/atom-beautify/blob/1498555b42591bbdcc1488b1ec02e9d787da0cc3/package.json#L188), there should always be at least one installation of ESLint found and this error message should never be reached.

src/beautifiers/eslint.coffee
```diff
@@ -14,21 +15,35 @@ module.exports = class ESLintFixer extends Beautifier
...
+              atom.notifications?.addFatalError needInstallMsg
+              resolve undefined
+        CLIEngine = eslint.CLIEngine
+        cli = new CLIEngine(fix: true, cwd: projectPath)
+        result = cli.executeOnText(text).results[0]
+        if not result.output
+          errMsg = result.messages[0].message
+          errLine = result.messages[0].line
+          errColumn = result.messages[0].column
+          @deprecate 'Eslint parse error: [' + errMsg + '] at line ' + errLine + ', column ' + errColumn
```
Glavin001, line 47: Is `@deprecate` appropriate to use here?

src/beautifiers/eslint.coffee
```diff
@@ -14,21 +15,35 @@ module.exports = class ESLintFixer extends Beautifier
...
+    return new @Promise((resolve, reject) =>
       editor = atom.workspace.getActiveTextEditor()
       filePath = editor.getPath()
       projectPath = atom.project.relativizePath(filePath)[0]
 
       result = null
-      allowUnsafeNewFunction ->
+      allowUnsafeNewFunction =>
         importPath = Path.join(projectPath, 'node_modules', 'eslint')
+        eslint = null
```
Glavin001, line 26: Tests should be added for these new cases.

PR #2258, opened 2018-11-08
Updates to Unibeautify branch

stevenzeck:
* Uses Winston for logging
* Adds php-cs-fixer beautifier
...

...

### Any other comments?

...

### Checklist


src/AtomBeautify.ts
```diff
@@ -1,203 +1,334 @@
...
-        const onWillSaveHandler = ({path: filePath}: {
-          path: string
-        }) => this.beautifyOnSaveHandler({
-          filePath: filePath
-        }, editor);
-        return this.subscriptions.add(editor.getBuffer().onWillSave(onWillSaveHandler as any));
-      });
+  // Retrieve the configuration options for Atom Beautify to display in Atom Settings
+  public get config() {
+    return Object.assign(Config, require("./options.json"));
```
Glavin001: `_.merge` is a recursive / deep merge of properties: https://lodash.com/docs/#merge

Using `Object.assign` is OK so long as `./options.json` does not contain a `general` property like `Config` does: https://github.com/Glavin001/atom-beautify/blob/3b4508e6fe1842e8058684a8edbd42f09bf9b8ba/src/config.ts#L2
I think this is true.

Glavin001: Should `./options.json` be committed to Git?
It is for current Atom-Beautify: https://github.com/Glavin001/atom-beautify/blob/master/src/options.json

src/AtomBeautify.ts
```diff
@@ -1,203 +1,334 @@
...
+    // TODO
+  }
 
-    private setScrollTop(editor: any, value: any): void {
-      const view = atom.views.getView(editor);
-      view.setScrollTop(value);
+  // Check if a language is configured to beautify on save
+  private isBeautifyOnSave(language: string): Boolean {
+    const languageConfig = this.baseConfig.languages;
+    if (languageConfig.hasOwnProperty(language)) {
```
Glavin001: This can be simplified to:

```diff
- if (languageConfig.hasOwnProperty(language)) {
-     return Boolean(_.get(languageConfig, language).beautify_on_save);
- }
- return false;
+ return Boolean(_.get(languageConfig, [language, "beautify_on_save"]));
```

See https://lodash.com/docs/#get

src/logger.ts
```diff
@@ -0,0 +1,59 @@
+import { format, TransformableInfo } from "logform";
+import * as path from "path";
+import * as winston from "winston";
+const { LEVEL, MESSAGE } = require("triple-beam");
```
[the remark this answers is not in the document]
stevenzeck: Where?  Don't see them here https://github.com/winstonjs/triple-beam

src/AtomBeautify.ts
stevenzeck: It can be, but I don't see any reason why it should.  The file is autogenerated.

src/logger.ts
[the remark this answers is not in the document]
Glavin001: They are not builtin. They are in the `@types/triple-beam` package I linked above.
See https://basarat.gitbooks.io/typescript/docs/types/@types.html for a brief description on `@types`.

src/AtomBeautify.ts
Glavin001: It doesn't need to be committed but it does need to be bundled into the package when pushed to Atom package registry. It should not be generated on the user's machine during installation or otherwise.

script/index.ts
```diff
@@ -52,39 +90,75 @@ function buildOptionsForLanguage(language: Language, beautifiers: String[]) {
...
+      order: index,
+      properties: beautifierProperties,
+    };
   });
+  return beautifierOptions;
 }
 
+function buildOptionsForBeautifier(beautifier: Beautifier) {
+  const beautifierOptions: AtomConfigRegistry = {};
+  if (!!beautifier.resolveConfig) {
```
Glavin001:
```suggestion
  if (beautifier.resolveConfig) {
```

stevenzeck: It doesn't return a boolean though, just Promise<ResolvedConfig>.

Glavin001: It's still truthy: https://developer.mozilla.org/en-US/docs/Glossary/truthy 

```
if (!!beautifier.resolveConfig) {
```

and

```
if (beautifier.resolveConfig) {
```

are effectively the same.

package.json
```diff
@@ -137,45 +137,64 @@
...
-    "cosmiconfig": "^4.0.0",
-    "eslint": "^4.19.1",
-    "js-beautify": "^1.7.5",
-    "prettier": "^1.11.1",
+    "@unibeautify/beautifier-black": "^0.1.1",
+    "@unibeautify/beautifier-brittany": "0.0.1",
+    "@unibeautify/beautifier-clang-format": "^0.2.0",
+    "@unibeautify/beautifier-csscomb": "^0.1.0",
+    "@unibeautify/beautifier-eslint": "^0.6.0",
+    "@unibeautify/beautifier-file": "^0.1.0",
```
stevenzeck: @Glavin001 what does this beautifier do?  Seems to have language support without any specified.

Glavin001: @stevenzeck The file beautifier is for generic options: 
- end of line
- end with newline

See https://github.com/Unibeautify/website/pull/162

It's useful to put as last beautifier in `beautifiers` array and then you can change the line endings and add/remove newline at end of file.

stevenzeck:
@Glavin001 pushed changes.

PR #2542, opened 2020-12-01
update unibeautify

UziTech:
Updates to unibeautify

- add all @unibeautify/* beautifiers
- use eslint instead of tslint since it is deprecated
- use github actions for testing instead of travis
- setup semantic-release for automated releases


UziTech:
@Glavin001 @stevenzeck I'm not sure if you want any of this but I thought this could help get the package back up to speed.

I haven't done much testing to make sure it actually works. Also some of the beautifiers might need to be updated to work with updated peer dependencies. or maybe we just need to include downgraded dependencies.

